PR #6, opened 2023-02-21
Update README.md

TonyGiorgio:


futurepaul:
heck yeah. I actually need to update the readme I don't use pnpm anymore

PR #12, opened 2023-04-05
Nostr tweaks

futurepaul:


futurepaul:
(this fixes the waitlist nostr view so it shows more of our posts, I was filtering too aggressively before)

PR #19, opened 2023-04-12
Get close to POC feature parity

futurepaul:
goal right now is to get all the info on the page and then refine / design from there. pretty happy with how many scenarios / use cases this covers already

TODO:
- [x] wire up scanning
- [x] figure out what I want to do about the background color
- [x] actually succeed at receiving would feel good

Screenshot 2023-04-12 at 6 47 38 PM

Screenshot 2023-04-12 at 6 47 32 PM


futurepaul:
here's the video tony

https://user-images.githubusercontent.com/543668/231920375-74f179a9-8235-4f2e-b5db-09e7817456a1.mov

futurepaul:
still feels a little cluttered / tight but I think it's progress.

the fee display makes a lot better use of the horizontal space on the "amount" section... hopefully it's not too much info in one spot

Screenshot 2023-04-15 at 11 19 45 AM

futurepaul:
now we actually pop up a modal when the send goes through, so that's nice.

also have added activity feeds for lightning, utxos, onchain

probably will merge this once I have polling on the receive page as well... getting closer to the end-to-end experience

PR #35, opened 2023-04-22
add instructions for linking

futurepaul:


benthecarman (approved):

justfile
```diff
@@ -0,0 +1,7 @@
+set dotenv-load := false
+
```
benthecarman, line 2: would be nice to have a run command too

benthecarman:
Doesn't seem to work for me

futurepaul:
did you use the new instruction here? https://github.com/MutinyWallet/mutiny-node/pull/394

and are your `mutiny-node` and `mutiny-web` folders siblings?

benthecarman:
> and are your mutiny-node and mutiny-web folders siblings?

they are both in `~/projects`

however `mutiny-wasm` is inside of mutiny-node, so idk if that would break something?

futurepaul:
no, that's the same for me

benthecarman (approved):

PR #36, opened 2023-04-22
Waila

futurepaul:


src/routes/Scanner.tsx
```diff
@@ -1,9 +1,48 @@
...
+    console.log("our network:", ourNetwork)
+
+    // TODO: "testnet" and "signet" are encoded the same I guess?
+    if (params.network === "testnet" || params.network === "signet") {
+        if (ourNetwork === "signet") {
+            // noop
+        }
+    } else if (params.network !== ourNetwork) {
+        return { ok: false, error: new Error(`Destination is for ${params.network} but you're on ${ourNetwork}`) }
+    }
```
benthecarman, line 31: We do these checks in the backend, probably okay for you to skip

futurepaul: If i do it here i fails in the scanner ui instead of pretending like it's valid until they send, is there a way i can ask the backend for help?

benthecarman: Okay that makes sense. Maybe keep this for now and I can add something to waila for this

PR #42, opened 2023-04-24
Update mutinynet RGS url

benthecarman:


futurepaul (approved):

benthecarman:
> This needs an update to mutiny-wasm in order to work

it was broken either way, this will eventually fix.

PR #66, opened 2023-04-28
More delete improvements

futurepaul:
hardcode network for now because network switching doesn't really work with delete shenanigans. will restore it in the "I do what I'm doing page" soon

### DIFF
--- a/src/components/DeleteEverything.tsx
+++ b/src/components/DeleteEverything.tsx
@@ -2,6 +2,8 @@ import { createSignal } from "solid-js";
 import { ConfirmDialog } from "~/components/Dialog";
 import { Button } from "~/components/layout";
 import { showToast } from "~/components/Toaster";
+import { useMegaStore } from "~/state/megaStore";
+import eify from "~/utils/eify";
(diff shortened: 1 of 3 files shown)


src/components/DeleteEverything.tsx
```diff
@@ -21,16 +22,24 @@ export function deleteDb(name: string) {
...
-        setConfirmOpen(false);
-        setConfirmLoading(false);
-        setTimeout(() => {
-            window.location.href = "/";
-        }, 3000);
+        try {
+            setConfirmLoading(true);
+            deleteDb("gossip")
+            deleteDb("wallet")
+            localStorage.clear();
```
benthecarman, line 29: Instead of this could instead import an empty json object into the node manager, that would clear it properly and if we change the indexed db stuff we won't need to remember to change it here.

futurepaul: oh dang that would be sweet

futurepaul:
used some of this logic for #84 with you suggestion

PR #70, opened 2023-04-29
Testing full clearing of node manager

TonyGiorgio:
Not complete and relies on latest mutiny-node branch.


futurepaul:
used some of this logic for #84

PR #120, opened 2023-05-13
Stop nodes in settings

TonyGiorgio:


futurepaul (approved):
works on my machine. probably need to do more to make this useful (like clear the sync interval or set some "stopped" state so the frontend doesn't misbehave). but this is good progress and we've agreed settings isn't made for everyone yet

PR #132, opened 2023-05-15
lightning details view

futurepaul:
caveats: the eyeballs don't work yet, the contacts aren't showing nice gradients for some reason, and I haven't done on-chain at all but I think I can punt on that for now because the json view still works in a pinch

edit: oh btw descriptions show if there's a description

| send         | receive   
|--------------|-----------|
| Screenshot 2023-05-15 at 12 25 17 PM | Screenshot 2023-05-15 at 12 25 01 PM |


TonyGiorgio:
I really like those modals, nice work. I would say though, I think it would be better if the eyeballs were just copy buttons instead. Typically there's no reason to "view" them, more likely they are need to copy them for some reason.

TonyGiorgio (approved):
looks good, minor changes can be in a follow up.

PR #140, opened 2023-05-16
Make imports await on file upload

TonyGiorgio:
Imports need to be awaited, but can't happen with `onload` so that's why a bigger change was needed. Dont ask me how it works but it works now. Otherwise state would not fully load.


futurepaul (approved):
I'm not great with hand-rolled promises either but it works on my machine, just restored nicely from my main test wallet to an incognito window

PR #141, opened 2023-05-17
Add Nostr wallet connect modal

benthecarman:
I just copy pasted the peer connection modal

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -26,6 +26,7 @@ export type MegaStore = [{
     has_backed_up: boolean,
     dismissed_restore_prompt: boolean,
     wallet_loading: boolean
+    nwc_enabled: boolean
 }, {
     fetchUserStatus(): Promise<UserStatus>;
     setupMutinyWallet(settings?: MutinyWalletSettingStrings): Promise<void>;
@@ -36,6 +37,7 @@ export type MegaStore = [{
     dismissRestorePrompt(): void;
     setHasBackedUp(): void;
     listTags(): Promise<MutinyTagItem[]>;
+    setNwc(enabled: boolean): void;
 }];
 
 export const Provider: ParentComponent = (props) => {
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
Might need to rebase. I think this is failing because package.lock wasn't updated here with the new mutiny version.

benthecarman:
> Might need to rebase. I think this is failing because package.lock wasn't updated here with the new mutiny version.

rebased

src/logic/mutinyWalletSetup.ts
```diff
@@ -87,12 +87,17 @@ export async function setupMutinyWallet(settings?: MutinyWalletSettingStrings):
...
     // If we don't have any nodes yet, create one
     if (!nodes.length) {
-        await mutinyWallet?.new_node()
+        await mutinyWallet?.new_node();
+        nodes = await mutinyWallet?.list_nodes();
     }
 
+    const firstNode = nodes[0] as string || "";
+
+    await mutinyWallet?.start_nostr_wallet_connect(firstNode);
```
TonyGiorgio, line 100: Maybe instead of having it auto run on startup we can make a "start nwc" button in the same settings area as the model. 

My concern is that it's experimental and no way to stop someone from stealing all your funds. I would rather we keep playing with this.

TonyGiorgio (changes requested):
Sorry hit the wrong button

src/logic/mutinyWalletSetup.ts
benthecarman: done

benthecarman:
Removed the auto startup, maybe @futurepaul can implement a button that enables start up

src/state/megaStore.tsx
```diff
@@ -136,6 +146,13 @@ export const Provider: ParentComponent = (props) => {
         },
         async listTags(): Promise<MutinyTagItem[]> {
             return state.mutiny_wallet?.get_tag_items() as MutinyTagItem[]
+        },
+        setNwc(enabled: boolean) {
+            localStorage.setItem("nwc_enabled", enabled.toString())
+            setState({ nwc_enabled: enabled })
+        },
+        async startNwc(mw: MutinyWallet): Promise<void> {
+
```
benthecarman, line 155: need to impl this

TonyGiorgio:
When I refresh after it is enabled, it does not start nwc

TonyGiorgio (approved):

PR #149, opened 2023-05-22
Real fees

futurepaul:
leaving ln blank rn because we don't have anything to put there

can revisit this for on-chain so we can make fees editable

also fixed lame bug in activity view that wasn't showing the correct sent amount

lightning | onchain
---|---
Screenshot 2023-05-22 at 12 06 03 PM | Screenshot 2023-05-22 at 12 06 10 PM


TonyGiorgio:
Tested it, looks good to me!

TonyGiorgio (approved):

PR #156, opened 2023-05-23
Swap to ln

futurepaul:


TonyGiorgio:
It says "no balance" when there's an unconfirmed balance. It should still be able to open channels with unconfirmed utxos.

TonyGiorgio:
Putting the exact balance you have as the channel open amount should error out properly similar to when you try to open a channel with with too much balance. The API should be returning an error.

TonyGiorgio:
After connecting, instead of returning the list for the user to select the same peer they put in the connection string, it should auto populate with that peer so they can simply click the open button.

futurepaul:
> Putting the exact balance you have as the channel open amount should error out properly similar to when you try to open a channel with with too much balance. The API should be returning an error.

which API will error?

futurepaul:
> It says "no balance" when there's an unconfirmed balance. It should still be able to open channels with unconfirmed utxos.

I didn't know that was possible?

TonyGiorgio:
> > Putting the exact balance you have as the channel open amount should error out properly similar to when you try to open a channel with with too much balance. The API should be returning an error.
> 
> which API will error?

The estimation api, which is why the fee is not appearing.

TonyGiorgio:
Okay I think this is pretty solid now. Nice work!

TonyGiorgio (approved):

PR #157, opened 2023-05-23
redshift coming soon

futurepaul:
big diff is because of prettier, all I did was wrap the view in a div that overlays the controls

Screenshot 2023-05-23 at 12 16 56 PM


TonyGiorgio:
Awesome, tested it and it looks good.

PR #161, opened 2023-05-24
lint extravaganza

futurepaul:
before:
Screenshot 2023-05-23 at 4 53 23 PM

after:
Screenshot 2023-05-23 at 7 04 16 PM

"how did it get this bad?" you might ask

well, I need to get linting into CI so I don't skip on it before a PR, so that's my story

also I got it set up better in my editor to show lints inline

two remaining errors are nostr-tools' fault

made some minor layout tweaks along the way trying to see if I can improve the amount keypad


TonyGiorgio (approved):
Every single line looks pretty nice work

PR #167, opened 2023-05-25
Catch and log lnurl auth errors

benthecarman:
Was testing some lnurl auth stuff and found that this was helpful


TonyGiorgio (commented):
If you've added console logging here for mutiny-core, does this need to be here again?

benthecarman:
Good point

PR #187, opened 2023-05-29
fee and min amount warnings

futurepaul:
warnings!

| | |
|--|--|
| Screenshot 2023-05-29 at 1 29 36 PM | Screenshot 2023-05-29 at 1 30 07 PM |
| Screenshot 2023-05-29 at 1 30 03 PM | Screenshot 2023-05-29 at 1 30 11 PM |

### DIFF
--- a/src/components/Amount.tsx
+++ b/src/components/Amount.tsx
@@ -41,7 +41,9 @@ export function AmountSmall(props: {
   return (
     <span class="font-light">
       {prettyPrintAmount(props.amountSats)}&nbsp;
-      <span class="text-sm">SATS</span>
+      <span class="text-sm">
+        {props.amountSats === 1 || props.amountSats === 1n ? "SAT" : "SATS"}
+      </span>
     </span>
-  )
+  );
 }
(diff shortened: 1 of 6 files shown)


TonyGiorgio:
I think this screen should show the fee indicator too. For one, on unified they might not know if it was on chain or lightning that was paid. Plus for normal lightning receives without a channel there is a 1 sat fee too so that should be shown. Needs to be clear before and after that a fee was paid so there's no surprises. 

Screenshot from 2023-05-29 16-56-28

TonyGiorgio:
This warning is appearing despite a channel already opened. 

Screenshot from 2023-05-29 16-58-58

futurepaul:
okay addressed the problem with the warning showing up and added more logic for whenever inbound is below request amount. also tweaked the language in the text below the qr code to warn the sender chooses in the unified case.

also added the explanation to the success screen

Screenshot 2023-05-30 at 3 08 48 PM

TonyGiorgio:
Screenshot from 2023-05-30 16-53-40

a 1 sat fee is taken out on the voltage side if a channel is already there. should we say that was a fee?

TonyGiorgio:
Other than that everything works great, tested a bunch of cases.

TonyGiorgio (approved):

PR #192, opened 2023-05-31
amount input switchable to usd

futurepaul:
Screenshot 2023-05-31 at 4 38 45 PM

### DIFF
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -1,7 +1,7 @@
-import { For, ParentComponent, Show, createMemo, createResource, createSignal } from "solid-js";
+import { For, ParentComponent, Show, createResource, createSignal, onMount } from "solid-js";
 import { Button } from "~/components/layout";
 import { useMegaStore } from "~/state/megaStore";
-import { satsToUsd } from "~/utils/conversions";
+import { satsToUsd, usdToSats } from "~/utils/conversions";
 import { Dialog } from "@kobalte/core";
 import close from "~/assets/icons/close.svg";
 import pencil from "~/assets/icons/pencil.svg";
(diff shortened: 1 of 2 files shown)


src/components/AmountEditable.tsx
```diff
@@ -204,48 +261,66 @@ export const AmountEditable: ParentComponent<{
...
+                onInput={handleSatsInput}
+              />
+              <label classList={{ "bg-pink-500": mode() === "fiat" }}>Fiat</label>
+              <pre>{displayFiat()} USD</pre>
+              <input
+                ref={(el) => (fiatInputRef = el)}
+                disabled={mode() === "sats"}
+                type="text"
+                class="bg-black text-white"
+                // class="opacity-0 absolute -z-10"
```
benthecarman, line 284: nit: can remove this i think

benthecarman (changes requested):
The fiat to sats conversion doesn't seem to work anymore

image

benthecarman:
Also not sure I love the way to switch between fiat and sats, normally you would just click the fiat number. I only really knew the swap button was the button for this because I was reviewing the PR

src/components/AmountEditable.tsx
futurepaul: thx

futurepaul:
okay updated with tap the amount to switch... also the price api seems to have been weird recently, can you try again?

btw you have to start on the home screen and then go to the receive screen to get price data... not exactly sure why. I checked and the same thing is true for the existing amount input screen

TonyGiorgio:
When I type in a few numbers, tap the switch and try typing more numbers, it does not work.

futurepaul:
> When I type in a few numbers, tap the switch and try typing more numbers, it does not work.

okay this was a very helpful comment I realized I can just always let the person keep typing in fiat view and just shift the decimal... before I was deleting anything the user typed past two decimals which is what you were seeing when you entered some sats then switched to fiat

TonyGiorgio:
Looks and works great!

benthecarman (approved):

PR #193, opened 2023-06-01
Bump to 0.3.6 & no more syncing for the frontend

TonyGiorgio:
Depends on a new cut of mutiny-wasm.

It's behaving much more fluid now. Excited about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mws",
-  "version": "0.3.5",
+  "version": "0.3.6",
   "license": "MIT",
   "scripts": {
     "dev": "solid-start dev",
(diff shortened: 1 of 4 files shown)


benthecarman:
Probably want to remove the syncing indicator on the on-chain balance too

TonyGiorgio:
> Probably want to remove the syncing indicator on the on-chain balance too. Can make @futurepaul do that tho

Handled that in here.

package.json
futurepaul, line 1: can you bump the mws package version as well. I'm keeping it in sync with the mutiny-wasm version

futurepaul (changes requested):
works on my machine, just would like the version bumped

futurepaul (approved):

PR #195, opened 2023-06-01
Send button now checks for insufficientFunds()

### LINKED ISSUE #97: Disable send button if insufficient funds
If send amount is > amount of funds in chosen send method, send button should be disabled

benalleng:
Depending on the source the user selected the send button checks their balance against the amountSats() sent to determine if the button should be disabled.

Closes #97


src/routes/Send.tsx
```diff
@@ -207,6 +207,15 @@ export default function Send() {
     setFieldDestination("");
   }
 
+  const insufficientFunds = () => {
+    if (source() === "onchain") {
+      return (state.balance?.confirmed ?? 0n) < amountSats()
```
futurepaul, line 212: should be the sum of confirmed + unconfirmed (I recently learned you can send from unconfirmed, changed my life)

futurepaul (changes requested):
nice improvement!

please squash your commits into a single commit and make sure to run prettier, there's a prettier config that prettier should pick up

benalleng:
With the introduction of #202 sending onchain now presents a clear error when the user has insufficient funds but lightning does not. Going to go ahead and make a new issue to address @benthecarman comment from element

> Would be confusing to enter an amount and can't figure out why I can't press the send button

> Yeah maybe putting some red error message under the amount saying insuffienct funds would make it better

benalleng:
Going to make a new PR, I messed up the rebase and now the history is totally borked

PR #198, opened 2023-06-01
lengthen interval, set mainnet channel size to 50k

futurepaul:
also fixes a namechange in invoice struct


TonyGiorgio:
Looks good, was just about to report the error around is_send.

PR #202, opened 2023-06-02
Fix amountbox

futurepaul:
this should fix the errors when scanning unified qrs / receive with no input... basically I was trying to focus the text input of the amount input even when the modal wasn't open, so it would throw an error

also caught a bug while doing this where the fee estimator for onchain was throwing errors that I wasn't catching, now added an error box possibly related to #195 stuff @benalleng


TonyGiorgio:
Works for me

TonyGiorgio (approved):

PR #204, opened 2023-06-02
Add force close balance to unconfirmed balance

benthecarman:


futurepaul (approved):
didn't run but looks good to me

PR #206, opened 2023-06-05
Send button now checks for insufficientFunds()

### LINKED ISSUE #97: Disable send button if insufficient funds

benalleng:
Closes #97 

initially opened a PR in #195 but i completely messed up the rebase, after several attempts to fix it i decided it would be easier to just close that PR

### DIFF
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -232,6 +232,24 @@ export default function Send() {
         setFieldDestination("");
     }
 
+    const insufficientFunds = () => {
+        if (source() === "onchain") {
+            return (
+                (state.balance?.confirmed ?? 0n) +
+                    (state.balance?.unconfirmed ?? 0n) <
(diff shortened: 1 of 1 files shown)


src/routes/Send.tsx
```diff
@@ -208,6 +208,16 @@ export default function Send() {
     setFieldDestination("");
   }
 
+  const insufficientFunds = () => {
+    if (source() === "onchain") {
+      return (state.balance?.confirmed ?? 0n) + (state.balance?.unconfirmed ?? 0n) < amountSats();
+    }
+    if (source() === "lightning") {
+      setError("We do not have enough balance to pay the given amount.");
```
futurepaul, line 216: this sets lightning to an error no matter what

futurepaul (changes requested):

PR #220, opened 2023-06-05
sync activity separately from regular sync

### LINKED ISSUE #212: No view all when nothing in activity

futurepaul:
this closes an IRL discussion where sync was causing lag on screens even that don't show activity. now activity only syncs on the home screen and activity screen

closes #212

this was built against mutiny-node master not sure if that's necessary tho actually


TonyGiorgio:
Seems to work! Will need to test it with a lot of data to see if it really works but should be better.

TonyGiorgio (approved):

PR #223, opened 2023-06-06
Add network subheading

### LINKED ISSUE #172: Header should say which network you're on if not mainnet

benalleng:
Closes #172

shows a network subheading when not on mainnet

image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -18,7 +18,19 @@ export default function App() {
         <SafeArea>
             <DefaultMain>
                 <header class="w-full flex justify-between items-center mt-4 mb-2">
-                    <img src={logo} class="h-10" alt="logo" />
+                    <div class="flex flex-col justify-center gap-2">
+                        <img src={logo} class="h-10" alt="logo" />
+                        <Show
+                            when={
+                                !state.wallet_loading &&
+                                state.mutiny_wallet?.get_network() !== "bitcoin"
+                            }
+                        >
+                            <div class="box-border px-2 py-1 -my-1 bg-white/70 rounded text-xs uppercase text-black w-fit">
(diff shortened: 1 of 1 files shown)


futurepaul (changes requested):
how about taking the styling from the `<Indicator />` component minus the pulse animation and putting it below the logo?

also the best way to get the network is `const network = state.mutiny_wallet?.get_network()`

and thirdly I think this should only show if on testnet / signet / regtest. so `<Show when={network !== "bitcoin"}>`

benalleng:
I had some difficulty setting the network as a variable as it would be called before the wallet was no longer state.wallet_loading so i went with just directly calling the get network after the `<Show />` conditions were met

benalleng:
Lastly any desire to make a non-flashing <Indicator /> for future use in this PR?

src/components/App.tsx
```diff
@@ -18,7 +18,19 @@ export default function App() {
         <SafeArea>
             <DefaultMain>
                 <header class="w-full flex justify-between items-center mt-4 mb-2">
-                    <img src={logo} class="h-10" alt="logo" />
+                    <div class="flex flex-col justify-center">
```
futurepaul, line 21:
```suggestion
                    <div class="flex flex-col justify-center gap-2">
```

futurepaul (changes requested):
sweet just one tiny thing

futurepaul:
> Lastly any desire to make a non-flashing `<Indicator />` for future use in this PR?

good thought but can hold off for now, we're about to have a wholistic figma ui as reference so that can be a time to revisit the "componentization" of some of this stuff (I don't think there's any Indicator in the new design)

futurepaul (approved):
thank you!

PR #224, opened 2023-06-06
check for browser features

futurepaul:


TonyGiorgio (approved):
Works well for me on the few private browsers I tested it on!

PR #225, opened 2023-06-06
Add dismissed_backup to sessionStorage

### LINKED ISSUE #86: Backup Dismissing
Currently when you dismiss the backup notification, it comes right back when you switch pages. This should ideally wait a little bit before showing again. Perhaps keep in memory until next reload should be fine without storing additional state?

benalleng:
Closes #86


futurepaul (approved):
tested and works great

thank you!

PR #229, opened 2023-06-08
Skip error log if no contacts

benthecarman:
We'd always get this error log if we didn't tag a contact, now it'll skip the log

`Something went wrong with contact creation, proceeding anyway`


futurepaul (approved):
I think we could get rid of all these console.errors in here they're just for debug purposes (I was using console.error at the time because my console was so spammy). also if you return from the if / else you don't need to return outside of it

I can make these changes if you don't want to tho they're all my fault, this works as-is

PR #232, opened 2023-06-09
Bump to 0.3.8

TonyGiorgio:


futurepaul:
I'm getting payment timed out errors but my ln balance goes down. Receiver doesn't seem to get paid.

TonyGiorgio:
> I'm getting payment timed out errors but my ln balance goes down. Receiver doesn't seem to get paid.

do the logs indicate that you've gotten an expired payment?

futurepaul:
not that I'm seeing... now I see one of the payments actually (shows up in the activity view) ... waiting if the second one comes through

TonyGiorgio:
> not that I'm seeing... now I see one of the payments actually (shows up in the activity view) ... waiting if the second one comes through

existing expired issue.

futurepaul (approved):
tests pass and I was able to make a send eventually

PR #233, opened 2023-06-09
don't show warnings if it's a swap

### LINKED ISSUE #209: The lightning opening fee warning is present on the swap to lightning screen
it should not be there on the swap screen, noticed it when entering a high amount.

futurepaul:
closes #209

### DIFF
--- a/src/components/AmountCard.tsx
+++ b/src/components/AmountCard.tsx
@@ -72,6 +72,7 @@ export function AmountCard(props: {
     initialOpen?: boolean;
     isAmountEditable?: boolean;
     setAmountSats?: (amount: bigint) => void;
+    skipWarnings?: boolean;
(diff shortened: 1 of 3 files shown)


src/components/AmountCard.tsx
```diff
@@ -72,6 +72,7 @@ export function AmountCard(props: {
     initialOpen?: boolean;
     isAmountEditable?: boolean;
     setAmountSats?: (amount: bigint) => void;
+    isSwap?: boolean;
```
benthecarman, line 75: Should just call this something like "skipWarnings", we have another reason to skip in the future

benthecarman (changes requested):

TonyGiorgio (approved):
Works well for me.

PR #236, opened 2023-06-11
Update a dependency

TonyGiorgio:


futurepaul (approved):
works on my machine

benthecarman (commented):
Idk if this will work for ios PWAs but lgtm

PR #240, opened 2023-06-12
Update mutiny-wasm and Add button for resync wallet

benthecarman:


futurepaul:
you gotta `pnpm install` after updating the dep in package.json or the lockfile won't be updated

futurepaul (approved):
nice. I clicked the button

futurepaul:
my tests pass as well

PR #241, opened 2023-06-12
detect if multiple tabs are open

futurepaul:


benthecarman (approved):
works for me! awesome job

PR #254, opened 2023-06-15
Send and swap max

### LINKED ISSUE #62: Send Max button
Screenshot from 2023-04-27 14-20-52

I cant tell how much I have to send when I am trying to withdraw all

### LINKED ISSUE #169: Open Max Channel
Users have to keep guessing what amount to open a channel and risk having tiny change if they dont get it right. I think we do this for redshift already so need to expose the API's for normal channels.

futurepaul:
needs https://github.com/MutinyWallet/mutiny-node/pull/596 merged in

closes #62
closes #169

### DIFF
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -232,17 +232,26 @@ export default function Send() {
         setFieldDestination("");
     }
 
+    const maxOnchain = createMemo(() => {
+        return (
+            (state.balance?.confirmed ?? 0n) +
+            (state.balance?.unconfirmed ?? 0n)
+        );
+    });
+
+    const isMax = () => {
+        if (source() === "onchain") {
+            return amountSats() === maxOnchain();
+        }
+    };
+
     const insufficientFunds = () => {
         if (source() === "onchain") {
-            return (
-                (state.balance?.confirmed ?? 0n) +
-                    (state.balance?.unconfirmed ?? 0n) <
-                amountSats()
-            );
+            return maxOnchain() < amountSats();
         }
         if (source() === "lightning") {
             return (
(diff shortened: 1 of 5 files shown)


src/routes/Send.tsx
```diff
@@ -260,6 +272,12 @@ export default function Send() {
             address()
         ) {
             setError(undefined);
+
+            // If max we want to use the sweep fee estimator
+            if (isMax()) {
+                return state.mutiny_wallet?.estimate_sweep_tx_fee(address()!);
+            }
```
benthecarman, line 279: should move this into the `try` block

src/routes/Swap.tsx
```diff
@@ -136,29 +122,57 @@ export default function Swap() {
...
-                    const new_channel = await state.mutiny_wallet?.open_channel(
-                        firstNode,
-                        undefined,
-                        amountSats()
-                    );
-
-                    setChannelOpenResult({ channel: new_channel });
+                if (isMax()) {
+                    console.log("sweeping all to channel");
+                    if (hasLsp()) {
```
benthecarman, line 131: instead of all the if statements around `hasLsp()`, would be cleaner to do something like

```javascript
let peer = undefined;
if (hasLsp()) {
	peer = selectedPeer();
}
```

benthecarman (changes requested):

src/routes/Swap.tsx
futurepaul: thank you

futurepaul:
Just rebased and tested both swap max and onchain send max

TonyGiorgio (approved):
Seems to work well for me!

benthecarman (approved):

PR #258, opened 2023-06-16
Enhances onClose() functionality of <AmountEditable />

### LINKED ISSUE #205: Close (x) button on Set Amount screen moves user to next step instead of returning to main screen
The X at the top right of the set amount screen doesn't behave as expected. I assume it was meant to close out of the receive flow and back to the main balance screen. Instead it progresses the user forward to the Receive Bitcoin screen.

To reproduce: From the main screen, tap Receive button. Enter an amount. Hit X at the top right.

### LINKED ISSUE #260: Closing AmountEditable still saves changes
I think the best UX would be where amounts are only saved if the user actively presses the set amount button

benalleng:
Closes #205, closes #260

the functionalitty for the send route remains where it keeps you on the send page without sending you forward

additionally the input is reset back to the initialAmountSats when onClose() so there is not a secret error where the UI would show a different amoun than the input

### DIFF
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -247,6 +247,20 @@ export const AmountEditable: ParentComponent<{
         }
     }
 
+    function handleClose() {
+        props.setAmountSats(BigInt(props.initialAmountSats));
+        setIsOpen(false);
+        setLocalSats(props.initialAmountSats);
+        setLocalFiat(
+            satsToUsd(
+                state.price,
+                parseInt(props.initialAmountSats || "0") || 0,
+                false
+            )
+        );
+        props.exitRoute && navigate(props.exitRoute);
+    }
+
     // What we're all here for in the first place: returning a value
     function handleSubmit(e: SubmitEvent | MouseEvent) {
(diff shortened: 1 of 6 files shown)


benalleng:
realized this isn't useable for the send route converting to draft

src/components/AmountEditable.tsx
```diff
@@ -320,12 +323,44 @@ export const AmountEditable: ParentComponent<{
                 <div class={DIALOG_POSITIONER}>
                     <Dialog.Content
                         class={DIALOG_CONTENT}
-                        onEscapeKeyDown={() => setIsOpen(false)}
+                        onEscapeKeyDown={() => {
```
futurepaul, line 326: looks like this logic is identical to the close button, could probably pull it out as a function that gets used in both?

futurepaul (changes requested):
tested a few flows and this all seems to be working correctly to me. thanks for tackling! just one little refactor

futurepaul (dismissed):
big win, thank you!

futurepaul (approved):

PR #266, opened 2023-06-19
add beta warning

futurepaul:
this shows once on load and doesn't show again after it's dismissed

Screenshot 2023-06-19 at 3 03 52 PM


TonyGiorgio (approved):
Seems to work well across all my devices.

PR #268, opened 2023-06-19
Only update local storage settings when saving settings

benthecarman:
Before we'd create a whole new instance of mutinywallet which could cause bad issues


futurepaul (approved):
works on my machine thank you

PR #269, opened 2023-06-19
Swap fixes

TonyGiorgio:
ChatGPT: Suspense and asynchronous operations: You are using Suspense and FullscreenLoader to handle asynchronous operations. If an asynchronous operation is ongoing when you navigate away, and then it completes and changes the state after you have navigated, this could potentially cause a re-render.

---

There was a time race that kept calling the fee estimator like 3-4 times in the span of navigating back to home. For some reason, doing it so fast causes it to call many times. Waiting on the success screen for a few seconds and then hitting nice will cause it to have a "not enough funds error" in the log. Then when clicking on the swap button causes a "dangit" error because not enough funds for the operation. 

This hopefully fixes all the weird transitions around swaps but there's an overall problem of nothing protecting the parent component from being remounted and another mutiny node manager from being instantiated again which is absolutely dangerous.

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -224,6 +224,10 @@ export const Provider: ParentComponent = (props) => {
         }
     };
 
+    onCleanup(() => {
+        console.warn('Parent Component is being unmounted!!!');
+    });
+
     // Fetch status from remote on load
     onMount(() => {
         // eslint-disable-next-line
--- a/src/components/BalanceBox.tsx
+++ b/src/components/BalanceBox.tsx
@@ -60,7 +60,7 @@ export default function BalanceBox(props: { loading?: boolean }) {
                 <Show when={!props.loading} fallback={<LoadingShimmer />}>
                     <div class="flex justify-between">
                         <Amount amountSats={totalOnchain()} showFiat />
-                        <Show when={!emptyBalance()}>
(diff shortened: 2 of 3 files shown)


src/state/megaStore.tsx
```diff
@@ -224,6 +224,10 @@ export const Provider: ParentComponent = (props) => {
         }
     };
 
+    onCleanup(() => {
+        console.warn('Parent Component is being unmounted!!!');
```
benthecarman, line 228: Maybe should put the current timestamp of this into local storage or something so we can see if it happened on a mobile app

TonyGiorgio: I don't think we do a export of all the data that might be in storage, just from what is relevant to mutiny-node. We could create something for the frontend to log and export but need to build that out.

benthecarman: oh yeah it wont

futurepaul: made an issue for this would be good for sluthing: https://github.com/MutinyWallet/mutiny-web/issues/271

futurepaul (approved):
worked on my machine for multiple swaps

TonyGiorgio (commented):
Sorry, my comment never posted. This is a problem @futurepaul

Issue #219, opened 2023-06-05
internationalization support


benalleng:
Eventually would be nice to switch over to an i18n framework or other internationalization system to allow for non English users

PR #275, opened 2023-06-20
Add i18n support

fernandoporazzi:
What about adding i18n support? I created an initial implementation and if you are happy with it, I could finish the translations setup and provide full translations for Portuguese and maybe Dutch.

Solves https://github.com/MutinyWallet/mutiny-web/issues/219


pnpm-lock.yaml
futurepaul, line 1: This megadiff is likely because of an outdated version of pnpm. Try updating pnpm to > 8.0 and rerunning `pnpm install`

src/components/App.tsx
```diff
@@ -11,9 +11,11 @@ import { useMegaStore } from "~/state/megaStore";
 import { Show } from "solid-js";
 import { ExternalLink } from "./layout/ExternalLink";
 import { BetaWarningModal } from "~/components/BetaWarningModal";
+import {useI18n} from "../i18n/context";
```
futurepaul, line 14: can use the `~` alias like this: `~/i18n/context` I know I'm not super consistent with it but it's nice especially for something like this that's gonna get imported all over the place

src/i18n/config.ts
```diff
@@ -0,0 +1,24 @@
+import i18next from 'i18next';
+import HttpApi from 'i18next-http-backend';
+import LanguageDetector from 'i18next-browser-languagedetector';
+const i18n = i18next
+  .use(HttpApi)
+  .use(LanguageDetector)
+  .init({
+    fallbackLng: 'en',
```
futurepaul, line 8: I get a warning in my browser that there's no `en-US` found... maybe that should be default?

src/root.tsx
```diff
@@ -12,12 +12,22 @@ import {
...
 import "./root.css";
 import { Provider as MegaStoreProvider } from "~/state/megaStore";
 import { Toaster } from "~/components/Toaster";
 import ErrorDisplay from "./components/ErrorDisplay";
+import { I18nProvider } from "./components/I18nProvider";
 
 export default function Root() {
+    const [loaded, setLoaded] = createSignal(false);
+
+    onMount(async () => {
```
futurepaul, line 26: I think you can clean this up a bit by putting the loading logic inside the provider. here's what I came up with:

```tsx
import { ParentComponent, Show, createResource } from "solid-js";
import { I18nContext } from "../i18n/context";
import i18next from "i18next";
import i18nConfig from "../i18n/config";

export const I18nProvider: ParentComponent = (props) => {
    const [i18nConfigured] = createResource(async () => {
        await i18nConfig;
        return true;
    });

    return (
        <Show when={i18nConfigured()}>
            <I18nContext.Provider value={i18next}>
                {props.children}
            </I18nContext.Provider>
        </Show>
    );
};
```

seems to get the same behavior and can just use

`<I18nProvider>` inside root without any props

(unless I'm missing some reason this def needs to be in root?)

src/components/I18nProvider.tsx
```diff
@@ -0,0 +1,14 @@
+import { JSXElement } from 'solid-js';
```
futurepaul, line 1: see comment on root, I rewrote this a little bit

futurepaul (changes requested):
Thank you so much! I've been dreading this and this looks like a great and simple solution.

A few minor adjustments and this should be good to merge.

src/i18n/config.ts
fernandoporazzi: Yes, I've seen that happening. That's because there isn't a file specifically for `en-US` or `en-GB` and so on. I am gonna do some reasearch and try to find a way to resolve all `en-SOME_ENGLISH_SPEAKING_COUNTRY` to just `en`. That's gonna be useful for other languages, such as `pt-BR` or `pt-PT`.

src/components/App.tsx
fernandoporazzi: Fixed. Marking this conversation as resolved.

src/root.tsx
fernandoporazzi: Oh, interesting solution. I implemented as you suggested and it worked like a charm. I am gonna mark this conversation as resolved as well.

src/i18n/config.ts
futurepaul: nice, works!

futurepaul (changes requested):
sweet this is looking good just one q about a dep

pnpm-lock.yaml
futurepaul, line 1: oh btw still need to fix your pnpm version and run `pnpm install`

fernandoporazzi: Yes, that's still pending. I am actually having a couple issues to get pnpm v8 to work 🥲

package.json
```diff
@@ -45,6 +45,10 @@
         "@solidjs/router": "^0.8.2",
         "@thisbeyond/solid-select": "^0.14.0",
         "class-variance-authority": "^0.4.0",
+        "i18next": "^22.5.1",
+        "i18next-browser-languagedetector": "^7.0.2",
+        "i18next-http-backend": "^2.2.1",
+        "i18next-xhr-backend": "^3.2.2",
```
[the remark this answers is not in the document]
fernandoporazzi: Nice catch. I guess we can remove that one!

pnpm-lock.yaml
futurepaul: i'm happy to do the install on my end if you don't mind me taking this over

fernandoporazzi: I tried it once more and I honestly don't know why I am stuck with pnpm v7.17. So if you fix the mega diff, I could push new commits with the translations.

futurepaul:
@fernandoporazzi alright should be good to go. No pressure on doing a whole translation at once, happy to do a little bit at a time / "DCA" into it

fernandoporazzi:
What about merging this PR and then opening a new one with the translations?

futurepaul:
let me know when you feel comfy with this, looks like you already added a bunch of translations... my fear is if this grows too big it'll be annoying to merge on top of master

fernandoporazzi:
I do feel like this could grow way too big and make the merge a bit complicated. What if we merge what's already here and then create separated PRs for each page of the app?

futurepaul:
> I do feel like this could grow way too big and make the merge a bit complicated. What if we merge what's already here and then create separated PRs for each page of the app?

yeah my thoughts exactly. I'll try and get this merged soon thank you so much for working on it!

futurepaul (approved):

futurepaul:
so stoked on this, thank you!

PR #277, opened 2023-06-20
Multi instance defense

futurepaul:
less noisy if you ignore the last commit, which is an attempt to catch all of these errors in the first place so none of this other defensive logic needs to be triggered

the actual very bad problem with the swap logic is that a potentially error-throwing function was being retriggered every time the state changed. the state changes when the balance goes down, which caused it to try and get the fee again.

IF this just resulted in them landing on the error page, then they'd be okay, because when they click "dangit" that reloads the page so they get a brand new `mutiny_wallet` instance and the old stuff is blown away

BUT because the error page unmounted the old mutiny_wallet without stopping it, AND then the swap logic auto-redirected to the home page, a new mutiny_wallet was getting initialized so now we had two mutiny_wallets.

enhancements in this pr:
1. stop mutiny on cleanup (when the parent component that holds all the state is unmounted, which should happen when we show the very bad error page)
2. if an existing session didn't clean up after itself, there's a sessionStorage var set so the page gets reloaded during setup (this makes setup take a bit longer, but should guard against having duplicate instances. this is less ideal than actually calling stop, but at this point in the object we don't have a reference to the mutiny_wallet object to stop it anymore)

a way to test this is to add a file called `Breaker.tsx` to the routes folder:

```tsx
import { createMemo, createSignal } from "solid-js";
import NavBar from "~/components/NavBar";
import {
    Button,
    DefaultMain,
    MutinyWalletGuard,
    SafeArea
} from "~/components/layout";
import { useMegaStore } from "~/state/megaStore";
import { useNavigate } from "solid-start";
import { BackLink } from "~/components/layout/BackLink";

const CHANNEL_FEE_ESTIMATE_ADDRESS =
    "bc1qf7546vg73ddsjznzq57z3e8jdn6gtw6au576j07kt6d9j7nz8mzsyn6lgf";

export default function Swap() {
    const [state, _actions] = useMegaStore();
    const navigate = useNavigate();

    createMemo;

    function breakMutiny() {
        setAmountSats(100000n);
        // comment this out if you want to stay on the error page
        navigate("/");
    }

    const [amountSats, setAmountSats] = createSignal(0n);

    const feeEstimate = createMemo(() => {
        if (amountSats() > 0n) {
            return state.mutiny_wallet?.estimate_tx_fee(
                CHANNEL_FEE_ESTIMATE_ADDRESS,
                amountSats(),
                undefined
            );
        }
        return undefined;
    });

    return (
        <MutinyWalletGuard>
            <SafeArea>
                <DefaultMain>
                    <pre>{feeEstimate()?.toString()}</pre>
                    <Button onClick={breakMutiny}>Break Mutiny</Button>
                    <BackLink />
                </DefaultMain>
                <NavBar activeTab="none" />
            </SafeArea>
        </MutinyWalletGuard>
    );
}

```

visit it at localhost:3420/breaker

(or you can go through the swap max logic again to retrigger manually)

### DIFF
--- a/src/logic/mutinyWalletSetup.ts
+++ b/src/logic/mutinyWalletSetup.ts
@@ -1,9 +1,7 @@
+/* @refresh reload */
+
 import initMutinyWallet, { MutinyWallet } from "@mutinywallet/mutiny-wasm";
 import initWaila from "@mutinywallet/waila-wasm";
-
-// export type MutinyWalletSettingStrings = {
-//     network?: string, proxy?: string, esplora?: string, rgs?: string, lsp?: string,
-// }
 
 export type Network = "bitcoin" | "testnet" | "regtest" | "signet";
 export type MutinyWalletSettingStrings = {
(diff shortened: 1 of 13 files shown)


src/logic/mutinyWalletSetup.ts
```diff
@@ -92,11 +90,23 @@ export async function checkForWasm() {
 export async function setupMutinyWallet(
     settings?: MutinyWalletSettingStrings
 ): Promise<MutinyWallet> {
+    // Ultimate defense against getting multiple instances of the wallet running.
+    // If we detect that the wallet has already been initialized in this session, we'll reload the page.
+    // A successful stop of the wallet in onCleanup will clear this flag
+    if (sessionStorage.getItem("MUTINY_WALLET_INITIALIZED")) {
```
benthecarman, line 96: This is really smart, nice work

benthecarman (approved):

TonyGiorgio (approved):
Did some heavy and fast testing on mobile and the browser. Nice work!

Issue #238, opened 2023-06-12
Workbox doesn't like index.html entry


futurepaul:
```
warnings
  Some items were passed to additionalManifestEntries without revisioning info. This is generally NOT safe. Learn more at https://bit.ly/wb-precache.
  - index.html
  ```

I believe this is related to this:

https://github.com/solidjs/solid-start/issues/518

benthecarman:
Can this be closed?

TonyGiorgio:
Yes we use index.html now

PR #280, opened 2023-06-20
cache after first load

futurepaul:
pwa is kind of hard to test locally so doing this PR so I get a deploy I can test

working on #238 mostly


futurepaul:
I didn't manage to get updates to work, too much stuff was breaking when I tried to update. or it would just show the new content without requiring the update. BUT. I did manage to get caching actually working so loads should be a lot faster now after install / first load

futurepaul:
closes #238

benthecarman:
Refreshing the preview seems much faster

futurepaul:
> Without manual updates, how would forcing an update work? Can you configure cache time or something?

this the `autoUpdate` behavior from vite-pwa: https://vite-pwa-org.netlify.app/guide/auto-update.html#automatic-reload

on further reading this might not be good because eventually we will want `prompt` and they warn upgrading from `autoUpdate` to prompt might be hard:

Screenshot 2023-06-21 at 2 14 46 PM

futurepaul:
changing this to draft while I try to figure out how to actually do `prompt` correctly

futurepaul:
my plan right now is to move away from solid-start and just use solid-js which can give us a single bundle so we can do atomic upgrades

PR #284, opened 2023-06-21
Twitter link preview card

### LINKED ISSUE #247: Add link preview cards
When you tweet app.mutinywallet.com it doesnt give you any nice link preview, would be nice to have something like that. Maybe for now we can have something about the waitlist and once launched we change it to a blurb about the wallet

benalleng:
Closes #247


benalleng:
Any idea if I can test this locally somehow?

benthecarman:
if the cloudflare preview showed up we could use that to test

futurepaul:
this is nice thank you! but let me get you a better image, hopefully in the next day or so

futurepaul:
twitter-card

@benalleng here you go. this fits their 2:1 aspect ratio for "summary card with large image": https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image

PR #285, opened 2023-06-21
fix: await scanner.start()

sectore:
All are not critical, but maybe worth to check/fix it:

- [x] `scanner.start()` is `async` (code https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L341)
- [x] No need to put `scanner` in a Signal, just keep a reference
- [x] Simplify `onCleanup` - no need to de-reference stuff, these are still local vars, which will be destroyed as soon as `Scanner` is out of scope and removed by its parent
- [x] Simplyfied `ref` registration by following https://www.solidjs.com/docs/latest/api#ref

### DIFF
--- a/src/components/Reader.tsx
+++ b/src/components/Reader.tsx
@@ -1,37 +1,32 @@
 import QrScanner from "qr-scanner";
-import { createSignal, onCleanup, onMount } from "solid-js";
+import { onCleanup, onMount } from "solid-js";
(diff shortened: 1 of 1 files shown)


futurepaul (approved):
nice catch thank you. tested in safari and chrome and working fine. appreciate it!

PR #286, opened 2023-06-21
Fix type lint errors

### LINKED ISSUE #148: Cleanup remaining lints
So we can add linting to CI

benalleng:
#148 

Not 100% sure that these type declarations are the correct way to fix these lints as I'm not totally familiar with handling Rust~Typescipt types

### Wontfix 
nostr-tools import SimplePool error

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -14,16 +14,13 @@ import {
     MutinyWalletSettingStrings,
     setupMutinyWallet
 } from "~/logic/mutinyWalletSetup";
-import {
-    MutinyBalance,
-    MutinyWallet,
-    ActivityItem as MutinyActivity
-} from "@mutinywallet/mutiny-wasm";
+import { MutinyBalance, MutinyWallet } from "@mutinywallet/mutiny-wasm";
 import { ParsedParams } from "~/routes/Scanner";
 import { MutinyTagItem } from "~/utils/tags";
 import { checkBrowserCompatibility } from "~/logic/browserCompatibility";
 import eify from "~/utils/eify";
 import { timeout } from "~/utils/timeout";
+import { ActivityItem } from "~/components/Activity";
 
 const MegaStoreContext = createContext<MegaStore>();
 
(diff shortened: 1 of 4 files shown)


src/components/waitlist/WaitlistAlreadyIn.tsx
```diff
@@ -9,7 +9,7 @@ const relayUrls = [
     "wss://eden.nostr.land"
 ];
 
-import { SimplePool } from "nostr-tools";
```
benalleng, line 12: This the the fix I am wondering the most about as it produces a inline lint in my nvim stating 
```
const pool = new SimplePool(); ~~This expression is not constructable.
```

futurepaul: yeah tbh I've been ignoring nostr-tools stuff because we're going to rip it out when we are done with the waitlist

benalleng: especially as 'nostr-tools' recommends the previous import

[nostr-tools readme file](https://github.com/nbd-wtf/nostr-tools/blob/b9a7f814aaa08a4b1cec705517b664390abd3f69/README.md?plain=1#L136)

wondering how else we can solve this lint

benalleng: going to go ahead and revert this one

src/state/megaStore.tsx
```diff
@@ -248,46 +245,46 @@ export const Provider: ParentComponent = (props) => {
         actions.fetchUserStatus().then((status) => {
             setState({ user_status: status });
 
+            function handleExisting() {
```
futurepaul, line 248: I don't normally like defining new functions inline but it def cleans this up

futurepaul (approved):
sweet this is a good improvement. I think eventually we might want to get these typescript definitions to part of the `mutiny-wasm` package so it's easier to stay in sync but haven't looked into how to do that yet

PR #287, opened 2023-06-21
add emergency kit and new error states

futurepaul:
relies on https://github.com/MutinyWallet/mutiny-node/pull/616

this covers a few bad error states.

1. if the node manager panics during setup, the error screen will include the "emergency kit" items of import / export, download logs, and delete everything, all of which will now work without a node manager present
2. on regular error screens (which shouldn't really happen anymore because we're catching all errors afaik) there's a link to a new `/emergencykit` route
3. in settings there's also a link to `/emergencykit`

basically, if someone is unable to load at all, they should be able to get to the emergency kit, and if they are able to load and navigate they can get to `/emergencykit` even if they don't have a url bar

### DIFF
--- a/src/components/ErrorDisplay.tsx
+++ b/src/components/ErrorDisplay.tsx
@@ -1,11 +1,13 @@
-import { Title } from "solid-start";
+import { A, Title } from "solid-start";
 import {
     Button,
     DefaultMain,
     LargeHeader,
+    NiceP,
     SafeArea,
     SmallHeader
 } from "~/components/layout";
+import { ExternalLink } from "./layout/ExternalLink";
 
 export default function ErrorDisplay(props: { error: Error }) {
     return (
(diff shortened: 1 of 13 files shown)


TonyGiorgio:
The emergency kit only loads on the home screen when it gets a panic. If the receive screen is refreshed and a panic occurs, it just shows a blank screen. Maybe that's okay?

TonyGiorgio (approved):
Works well for me now. Some pages won't have this but it'll be clear when they go to settings to investigate further that something is messed up. Should help troubleshoot more.

src/components/ErrorDisplay.tsx
```diff
@@ -18,6 +19,14 @@ export default function ErrorDisplay(props: { error: Error }) {
                     <span class="font-bold">{props.error.name}</span>:{" "}
                     {props.error.message}
                 </p>
+                <NiceP>
+                    Try reloading this page or clicking the "Dangit" button. If
+                    you keep having problems, reach out to us for support.
```
benthecarman, line 24: should add a link to reach out to us

TonyGiorgio (approved):
Works great

PR #300, opened 2023-06-26
fix setup error precendence

### LINKED ISSUE #296: Browser incompatibility screen is replaced by emergency screen
When an error relating to browser incompatibility occurs, it is flashed quickly and then taken over by the emergency screen. It still says the same error in the dialog, but it loses all the information about browsers.

futurepaul:
closes #296

### DIFF
--- a/src/components/SetupErrorDisplay.tsx
+++ b/src/components/SetupErrorDisplay.tsx
@@ -28,17 +28,20 @@ function ErrorFooter() {
     );
 }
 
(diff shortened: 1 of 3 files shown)


TonyGiorgio (approved):
I tested on my phone with JIT turned off and I tested when the phone should load the emergency page when stalling out. Not sure about other crashes triggering but from what I can easily test, looks good to me.

PR #307, opened 2023-06-29
NWC updates

benthecarman:
Changes needed after https://github.com/MutinyWallet/mutiny-node/pull/632

For now just creates a profile if one is not available when viewing the uri


futurepaul:
closing in favor of #309

PR #308, opened 2023-06-29
settings reorg

futurepaul:
Screenshot 2023-06-29 at 10 20 22 AM

a good chunk of the work toward the new settings stuff in figma, basic goal was to make a place to land the new nwc stuff, so now I'm ready for that


TonyGiorgio:
- Lightning channels page is completely blank if there is not a channel. Should probably say something. 

Otherwise super clean!

futurepaul:
> * Lightning channels page is completely blank if there is not a channel. Should probably say something.

good catch

benthecarman:
this is sexy af

benthecarman:
`Backup` being red makes it look like a scary button even thought it is quite the opposite. I feel like the debug tools should be red so those look scary

TonyGiorgio:
"Admin Page" would be a better name than "Kitchen Sink"

TonyGiorgio (approved):

Issue #267, opened 2023-06-19
Trim whitespace prefix/suffix from inputs


TonyGiorgio:
Some copy paste jobs might have new lines or white space at the beginning or end of an address or invoice. Hard to tell what's wrong in the input box, we should just trim these correctly.

PR #310, opened 2023-06-29
Trim on paste

bennyhodl:
had 10 minutes to contribute to something, closes https://github.com/MutinyWallet/mutiny-web/issues/267


futurepaul (changes requested):
thank you! this is a strict improvement, but also there's a similar but not identical `handlePaste` fn in `Send.tsx`

futurepaul (approved):
works great, thank you!

PR #311, opened 2023-06-29
qr redesign

futurepaul:
not pixel perfect with the design but it gets a lot of the way there and I think it's pretty. can land the rest of the tweaks at a later date

Screenshot 2023-06-29 at 6 15 35 PM

Screenshot 2023-06-29 at 6 21 35 PM
Screenshot 2023-06-29 at 6 16 55 PM


TonyGiorgio (approved):
Feels great!

PR #312, opened 2023-07-01
Add setting for auth URL

TonyGiorgio:
Relies on mutiny-node `master`. Needs next release.


futurepaul:
this was cherry-picked as part of #313

PR #317, opened 2023-07-07
add loading progress bar

futurepaul:
it's not exactly beautiful but I think this could be helpful for diagnosing where someone is "stuck" in loading... I couldn't find any obvious way to hook into the wasm load progress, so instead I just made checkpoints for setup to say how far it's gotten

also this defers the waila stuff to the scanner (so only loads when you pull up send or scanner) which should make the home screen load faster by itself

https://github.com/MutinyWallet/mutiny-web/assets/543668/9d2231f4-650e-4bf4-9083-1e6e0c160c69

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -12,19 +12,29 @@ import {
 import { createStore, reconcile } from "solid-js/store";
 import {
     MutinyWalletSettingStrings,
+    doubleInitDefense,
+    initializeWasm,
     setupMutinyWallet
 } from "~/logic/mutinyWalletSetup";
 import { MutinyBalance, MutinyWallet } from "@mutinywallet/mutiny-wasm";
-import { ParsedParams } from "~/routes/Scanner";
 import { MutinyTagItem } from "~/utils/tags";
 import { checkBrowserCompatibility } from "~/logic/browserCompatibility";
 import eify from "~/utils/eify";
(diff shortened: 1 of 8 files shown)


src/state/megaStore.tsx
```diff
@@ -135,14 +147,23 @@ export const Provider: ParentComponent = (props) => {
...
+                    wallet_loading: true,
+                    load_stage: "checking_double_init"
+                });
+                await doubleInitDefense();
+                setState({ load_stage: "downloading" });
+                await initializeWasm();
+                setState({ load_stage: "setup" });
                 const mutinyWallet = await setupMutinyWallet(settings);
                 // Get balance optimistically
-                const balance = await mutinyWallet.get_balance();
```
benthecarman, line 141: is removing this get balance intended?

futurepaul: oh I meant to ask about this... is this helping anything? i remove it and it doesn't seem to change much?

futurepaul:
Screenshot 2023-07-10 at 6 46 53 PM

Screenshot 2023-07-10 at 6 47 08 PM

benthecarman (approved):
This is great

src/state/megaStore.tsx
futurepaul: fixed (I moved it down but yeah we do need it)

PR #318, opened 2023-07-07
encrypt / decrypt

futurepaul:
alright this is looking good imo. I tested setting a password, removing the password, exporting with / without encryption

### DIFF
--- a/src/logic/mutinyWalletSetup.ts
+++ b/src/logic/mutinyWalletSetup.ts
@@ -106,7 +106,8 @@ export async function checkForWasm() {
 }
 
 export async function setupMutinyWallet(
-    settings?: MutinyWalletSettingStrings
+    settings?: MutinyWalletSettingStrings,
+    password?: string
 ): Promise<MutinyWallet> {
     // Ultimate defense against getting multiple instances of the wallet running.
     // If we detect that the wallet has already been initialized in this session, we'll reload the page.
@@ -138,7 +139,7 @@ export async function setupMutinyWallet(
     console.log("Using subscriptions address", subscriptions);
     const mutinyWallet = await new MutinyWallet(
         // Password
-        "",
+        password ? password : undefined,
         // Mnemonic
         undefined,
         proxy,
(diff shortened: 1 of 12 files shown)


futurepaul:
- [ ] make it so you can get rid of the password
- [ ] make sure it works with export

futurepaul:
getting an error when I try to encrypt with no existing password. pretty sure it's not my fault, was working before

Screenshot 2023-07-08 at 6 47 40 PM

src/routes/settings/Encrypt.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const [store, _actions] = useMegaStore();
+    const [error, setError] = createSignal<Error>();
+    const [loading, setLoading] = createSignal(false);
+
+    const [_encryptPasswordForm, { Form, Field }] =
+        createForm<EncryptPasswordForm>({
+            initialValues: {
+                existingPassword: "",
+                password: "",
+                confirmPassword: ""
```
TonyGiorgio, line 37: Use `undefined`

src/logic/mutinyWalletSetup.ts
```diff
@@ -138,7 +140,7 @@ export async function setupMutinyWallet(
     console.log("Using subscriptions address", subscriptions);
     const mutinyWallet = await new MutinyWallet(
         // Password
-        "",
+        password ? password : "",
```
TonyGiorgio, line 143: `undefined`

TonyGiorgio:
Existing password at the top should be removed if they're coming in from the "backup notification" flow. We know they're not going to have an existing password. 

Screenshot from 2023-07-10 14-19-13

TonyGiorgio:
We'll probably have some users that forget their password. How should we handle that? Take them to the emergency page? 

Screenshot from 2023-07-10 14-22-35

futurepaul:
> Existing password at the top should be removed if they're coming in from the "backup notification" flow. We know they're not going to have an existing password.

I agree this would be nice but requires a decent bit of logic to have a "referral page" so maybe just make an issue for now

futurepaul:
Screenshot 2023-07-10 at 3 47 30 PM

TonyGiorgio (approved):

PR #319, opened 2023-07-08
fix: i18n broken when deployed

futurepaul:
for some reason the `i18next` http package works fine on local host but not when deployed... probably because we deploy static

for now just going to import all the translations... maybe would be bad if we had a million but should be fine for now until we can find a way to do this lazy-loaded


benthecarman (approved):

TonyGiorgio (approved):

src/i18n/config.ts
```diff
@@ -1,26 +1,39 @@
...
-      order: ['querystring', 'navigator', 'htmlTag'],
-      lookupQuerystring: 'lang',
+import { use } from "i18next";
+// FIXME: this doesn't work when deployed
+// import HttpApi from 'i18next-http-backend';
+import LanguageDetector from "i18next-browser-languagedetector";
+import en from "~/i18n/en/translations";
+import pt from "~/i18n/pt/translations";
+
+const i18n = use(LanguageDetector).init(
```
benalleng, line 8: Not sure if it's necessary but you didn't call i18next here
```
const i18n = i18next.use(LanguageDetector).init(
```

benalleng (commented):
Linked in from discord, @ethanTuttle pointed out to me that the backed version wasn't functioning as expected. Do you think deploying it as an imported package as done in [fedimint/#24](https://github.com/fedimint/ui/pull/24) would have a better chance? It does use react so it might not transition well in concept

PR #325, opened 2023-07-11
remove waitlist

futurepaul:
felt pretty straightforward not seeing any gotchas


.env.example
```diff
@@ -11,6 +11,5 @@ VITE_PROXY="wss://p.mutinywallet.com"
 VITE_ESPLORA="https://mutinynet.com/api"
 VITE_LSP="https://signet-lsp.mutinywallet.com"
 VITE_RGS="https://rgs.mutinynet.com/snapshot/"
-VITE_SELFHOSTED="true"
```
benthecarman, line 14: we probably wanna keep this for free mutiny+, no?

src/routes/index.tsx
```diff
@@ -14,15 +12,9 @@ export default function Home() {
             <Match when={state.setup_error}>
                 <SetupErrorDisplay initialError={state.setup_error!} />
             </Match>
-            <Match when={state.user_status === "approved"}>
+            <Match when={true}>
```
benthecarman, line 15: maybe should just be `when={!state.setup_error}`

.env.example
futurepaul: yeah I thought about that but we don't have it wired up rn... guess I can leave in the env

PR #327, opened 2023-07-12
async activity and nwc refresh

futurepaul:
this is yet another attempt to make activity sync non-blocking of the overall page load

seems to be working fine on my machine but I don't have enough activity to cause laggy loads

also I made it so pendingnwc updates the same way


TonyGiorgio:
Works for me!

TonyGiorgio (approved):

PR #328, opened 2023-07-12
handle lsp errors

futurepaul:
we display the actual error returned from failing to generate an invoice

EXCEPT I'm handling lsp errors specifically

(more specific error cases can be handled just starting with these)

then, if we failed to get an invoice, we try to get an onchain address

and we just show the onchain address no unified or lightning stuff (and no method picker)

Screenshot 2023-07-12 at 2 10 01 PM


src/logic/errorDispatch.ts
```diff
@@ -0,0 +1,68 @@
...
+    | "The given node pubkey is invalid."
+    | "Failed to get the bitcoin price."
+    | "Satoshi amount is invalid"
+    | "Failed to execute a dlc function"
+    | "Failed to execute a wasm_bindgen function"
+    | "Invalid Arguments were given"
+    | "Incorrect password entered."
+    | "Unknown Error";
+
+export function matchError(e: unknown): Error {
```
futurepaul, line 49: I didn't really need to do all this fanciness but this is how eventually we'll be able to dispatch to translations for errors

TonyGiorgio (approved):

Issue #348, opened 2023-07-14
Should say `Expired` not `Expires`


suhailsaqan:
<img width="565" alt="Screenshot 2023-07-13 at 8 25 00 PM" src="https://github.com/MutinyWallet/mutiny-web/assets/43693074/01cae8fe-35c9-438b-99fb-edcb0cc0672c">

PR #349, opened 2023-07-14
`Expired` not `Expires` for expired items

suhailsaqan:


suhailsaqan:
Closes #348

benthecarman (approved):
We should be filtering these out on the backend but if they do show up we should have correct wording

futurepaul (changes requested):
thanks for spotting this!

need to divide  Date.now() by 1000, also I think the format string is broken. I think I'd prefer an utility function outside the component:

```
function formatExpiration(expiration?: bigint) {
    if (!expiration) {
        return "Unknown expiration";
    }

    if (expiration <= Date.now() / 1000) {
        return "Expired";
    }

    return `Expires ${timeAgo(expiration)}`;
}
```

futurepaul (approved):
sweet, thanks!

PR #374, opened 2023-07-17
move import / export to admin page

### LINKED ISSUE #371: Move import and export state buttons to admin page
Having these in the emergency kit makes users think they should be using this but seems to be breaking things for users. Probably best to hide this from reach.

futurepaul:
closes #371 

also reworded the emergency kit

Screenshot 2023-07-17 at 11 26 40 AM


TonyGiorgio:
We need it on the emergency page when a storage error pops up.

-------- Original Message --------

futurepaul:
closing for now we can revisit @benthecarman lemme know if you still wan this

Issue #290, opened 2023-06-22
Do some Static Rendering so we get some minimal page for JS-less browsers


futurepaul:
especially useful for rendering meta tags... right now we basically ship a blank html page + a script tag

Possibly want to use this... https://github.com/solidjs/solid-start/tree/main/packages/start-static the readme isn't very encouraging though lol

the main goal would be to find a way to have only a single page worried about "am I being rendered" on the server so we don't have to pepper that logic all over the app

benalleng:
One point of confusion for rendering these meta tags is how some metadata is shown when googling.
Both the `description` and `twitter:description` are displayed
![image](https://github.com/MutinyWallet/mutiny-web/assets/108441023/3a05b2c3-6991-4c81-93ca-2818a9363892)

futurepaul:
yeah google does a minor amount of js afaik

futurepaul:
looks like this could be solved by this PR: https://github.com/solidjs/solid-start/pull/810/files

PR #376, opened 2023-07-18
Android

TonyGiorgio:
As far as I can tell, fully functioning as is. 

TODO:
- [x] Play with version settings
- [x] App names and icons
- [x] Get a CI build process going
- [x] Fix clipboard and camera

Screenshot from 2023-07-17 23-01-34


package.json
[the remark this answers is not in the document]
TonyGiorgio: Camera didn't work unless it was in there, I can try to remove some or see what can be moved to devDependencies. We might want `core` for their api's around checking what platform it is on and stuff. We can do conditional things like recommend the app store for greater amounts and stuff.

benthecarman:
Will we have to make builds for updates will things always auto deploy like web?

TonyGiorgio:
> Will we have to make builds for updates will things always auto deploy like web?

I think I can automate most of it

TonyGiorgio:
I mostly have auto builds going on.

I think we have to do something about the camera and clipboard. Those are the only two things that don't work on my phone. 

https://capacitorjs.com/docs/apis/clipboard
https://capacitorjs.com/docs/apis/camera

TonyGiorgio:
I had to do a hacky script to get `<head>` tags showing up in the generated `index.html` file so that capacitor can inject their stuff into it. But that allowed me to get the clipboard plugin to work. Swapped out all clipboard checks with an `if` statement depending on if it's native or not.

The tricky one is going to be the camera. I think it'll have to be some default ugly camera popping up so we might lose our UI overlay over the camera, at least on native. And I don't know if we can stream frames over to our QR code scanner or not. This might be tough and might need @futurepaul's help.

README.md
```diff
@@ -56,3 +56,19 @@ To revert back and use the remote version of mutiny-wasm:
 ```
 just remote
 ```
+
+## Android
+
+### How to Deploy
```
benthecarman, line 62: should also add build instructions for testing

android/app/src/main/res/mipmap-xxxhdpi/ic_launcher_foreground.png
futurepaul, line 1: whoever thought it was smart for devs to have to make 5000 variants of the icon is no friend of mine

android/app/src/main/res/values/ic_launcher_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="ic_launcher_background">#FFFFFF</color>
```
futurepaul, line 3: just guessing this is the splash screen during launch? could change to `#171717` instead of white

insertHead.js
futurepaul, line 1: another reason we need a better way to edit the index.html (https://github.com/MutinyWallet/mutiny-web/issues/290)

TonyGiorgio:
The tools handle that.

-------- Original Message --------

futurepaul (approved):
works on my android, nothing seems off about the code... I only skimmed some of the big boilerplate stuff obvi.

also checked for regressions on the web version, camera and clipboard seem to be working good!

android/app/src/main/res/values/ic_launcher_background.xml
TonyGiorgio: didn't affect anything, i don't care if it's just white for now.

TonyGiorgio:
Rebased. Doesn't work with the new send / receive screens. I'll have to troubleshoot it this weekend.

TonyGiorgio:
Seems to just be an emulator problem.

PR #381, opened 2023-07-19
manual chunking

futurepaul:
experimenting with manual chunks for build so we don't have all these little pieces of js

https://rollupjs.org/configuration-options/#output-manualchunks

problem: it doesn't work

Screenshot 2023-07-19 at 2 56 45 PM


futurepaul:
I think the move will be to move away from solid-start and just do a regular solid-js build into one nice single bundle

PR #388, opened 2023-07-21
Initial transition to translation keys

benalleng:
first pass at transitioning over to i18n.t('namespace.key') for all copy.

I took a "componentization" approach so as to try and make the typescript file easier to read, let me know if you think you'd like a different approach with how the keys should be organized.

It is not perfect at the moment just a start


futurepaul (approved):
great, thank you! sorry for the slow review

PR #315, opened 2023-07-04
Send and Receive success screens

benalleng:
Adds new success Screens for Send and Receive as well as making some adjustments to the functionality of the Amount.tsx component to better handle its usecases

### WONTFIX:

- only adds an individual `<AmountFiat />` component for use in the swap route and does not separate all instances of `<Amount />` into its two parts as we are intending to do later on
- error hints or error help needs to be added for each error code sent from mutiny-node
- confirmation time estimates need to be added as functionality from mutiny-node
- transaction fee help copy still needs to be send-type-correct in other words all transaction fee help returns the same message at the moment
- transaction details screens need to be implemented, unless the link directs to a mempool.space transaction all view transaction links are hidden

PR #392, opened 2023-07-24
Simplify Amount.tsx and separate amountsats and amountfiat

benalleng:
builds on the discussion from #315 to separate and simplify the amount component throughout the codebase


src/components/Amount.tsx
```diff
@@ -89,34 +66,35 @@ export function Amount(props: {
...
+    classes?: string;
+    denominationSize?: "sm" | "lg" | "xl";
+}) {
+    const [state, _] = useMegaStore();
+
+    const amountInUsd = () =>
+        satsToUsd(state.price, Number(props.amountSats) || 0, true);
+
+    return (
+        <div class={`flex flex-col gap-1 ${props.classes}`}>
```
futurepaul, line 85: sorry :(

src/components/Amount.tsx
```diff
@@ -11,47 +11,23 @@ function prettyPrintAmount(n?: number | bigint): string {
...
     return (
-        <div
-            class="flex flex-col gap-1"
-            classList={{
-                "items-start": props.align === "left",
-                "items-center": props.align === "center",
-                "items-end": props.align === "right"
-            }}
-        >
+        <div class={`flex flex-col gap-1 ${props.classes}`}>
```
futurepaul, line 22: this string interpolation sadly doesn't work. it works _sometimes_ with tailwind, but not often enough to be reliable (due to the compile-time step of tailwind, basically it could fail to compile a variant and so at runtime you don't actually have all options available)

that's why I end up using `classList` a lot: https://www.solidjs.com/docs/latest/api#classlist

futurepaul (changes requested):
overall this feels really good, thank you for tackling. unfortunately that `props.classes` trick won't work

src/components/Amount.tsx
benalleng: bummer ok, i'll make this change

benalleng: I went ahead and moved all classes up one level so they are not passing the classes as props they just exist as the parent elements of amountsats and amountfiat now.

src/routes/Swap.tsx
```diff
@@ -304,13 +304,14 @@ export default function Swap() {
...
                                         balance
                                     </p>
-                                    <AmountFiat
-                                        amountSats={
-                                            channelOpenResult()?.channel
-                                                ?.balance
-                                        }
-                                        classes="text-sm text-center"
-                                    />
+                                    <div class="text-sm tet-center text-white/70">
```
futurepaul, line 307: text-center

futurepaul (changes requested):
just one typo I saw but otherwise looks good

futurepaul (approved):
nice, thank you. I tried this out on top of the type checking branch and it's all clean

PR #393, opened 2023-07-25
Close dialog on select receive flavor

benthecarman:
Will now close the dialog when you select an option from this screen instead of having the user need to click the X afterwards
image


futurepaul (approved):
works on my machine. I thought it would be trivial to make the ui actually reflect the change before closing the modal... but it wasn't. so will merge as-is and will try and revisit soon

PR #401, opened 2023-07-26
Use delete function for deleting while active

TonyGiorgio:
Depends on https://github.com/MutinyWallet/mutiny-node/pull/671

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -194,15 +194,14 @@ export const Provider: ParentComponent = (props) => {
         },
         async deleteMutinyWallet(): Promise<void> {
             try {
-                if (state.mutiny_wallet) {
-                    await state.mutiny_wallet?.stop();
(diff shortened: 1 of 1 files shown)


benthecarman:
oh i made #402 which handles the case where we cant init a wallet

TonyGiorgio:
> oh i made #402 which handles the case where we cant init a wallet

You didn't mention there was a PR already when I asked why it didn't work :man_facepalming: 

Either way, I think yours doesn't work. 
1. The state just got cleared of mutiny wallet
2. The code that calls this delete method already checks if it's init'd or not and calls `import_json` otherwise. So it's redundant here in the first place.

benthecarman:
> You didn't mention there was a PR already when I asked why it didn't work man_facepalming

I started making it after I said that lol

futurepaul:
cherry-picked to #406

PR #405, opened 2023-07-27
Add scorer api

TonyGiorgio:
Depends on latest release

PR #406, opened 2023-07-27
Update to 0.4.5

futurepaul:
note: I have a version of delete everything (the one that is in the emergency kit) that won't call the new delete function because it's assuming there's not mutiny_wallet loaded, or mutiny_wallet is in a bad state...

(includes #401 and #405 cherry-picked)


TonyGiorgio:
I just added the env variables in cloudflare for the scorer, so it's not reflected here but everything else seems to work properly.

package.json
```diff
@@ -44,7 +44,7 @@
         "@kobalte/tailwindcss": "^0.5.0",
         "@modular-forms/solid": "^0.18.0",
         "@mutinywallet/barcode-scanner": "5.0.0-beta.3",
-        "@mutinywallet/mutiny-wasm": "0.4.4",
+        "@mutinywallet/mutiny-wasm": "0.4.5",
```
TonyGiorgio, line 47: Needs updating for the package versioning too

futurepaul: nice catch

TonyGiorgio (changes requested):
oh actually so all the github actions for the android builds need to be updated too now for each env change: 

https://github.com/MutinyWallet/mutiny-web/blob/7c9c99208464aa923bc765295bfec8240aade8ab/.github/workflows/android-build.yml#L63-L71

.github/workflows/android-build.yml
```diff
@@ -69,6 +69,7 @@ jobs:
           VITE_AUTH: https://auth-staging.mutinywallet.com
           VITE_SUBSCRIPTIONS: https://subscriptions-staging.mutinywallet.com
           VITE_STORAGE: https://storage-staging.mutinywallet.com
+          VITE_SCORER: https://scorer-staging.mutinywallet.com
```
TonyGiorgio, line 72: this is missing too: 

```
VITE_FEEDBACK="https://feedback-staging.mutinywallet.com"
```

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #407, opened 2023-07-27
Test restore

futurepaul:
A few things ended up snowballing into one big pr but I tried to separate it out into logical commits

1. fixed up the restore test so that it doesn't try to run multiple restores in parallel, and also so it cleans up after itself
2. added a playwright workflow to CI. right now it only does one platform (chrome mobile). can probably add firefox. still can run ff, chrome, and safari locally.
3. added a check-types command that actually runs the typescript type checker (we'd only been running lints before now, big oversight!)
4. added a code-quality workflow that checks for formatting, lint errors, and typecheck errors
5. fixed some non-critical eslint stuff that had been bothering me
6. ran `npx cap sync` because some deps had updated


.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,52 @@
+name: Code Quality Check
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
```
TonyGiorgio, line 6: should be master and prod

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,66 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
```
TonyGiorgio, line 6: should be master and prod

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #412, opened 2023-07-28
Sync nostr contacts

benthecarman:
Requires https://github.com/MutinyWallet/mutiny-node/pull/681

Very ugly but functionality is here


futurepaul:
https://github.com/MutinyWallet/mutiny-web/assets/543668/30dc87c1-a794-4170-984f-085529caa41a

okay finally starting to come together.

todo:
- [ ] have to figure out a clean way to poll for new events without clobbering the list every time (rn I don't poll at all)
- [ ] ideally can use the contact list from the wallet (right now I fetch it based on the npub which is a good amount of the load time
- [ ] loading state
- [ ] I think this would be a nice feature for someone who doesn't even want to use the wallet, but rn you can't get to activity screen unless you have wallet activity
- [ ] figure out what additional data we want to display on expand: probably the note that was zapped? (maybe even preview the note before expand to save a click)
- [ ] we had some ideas about collapsing zaps if like one influencer is getting mega zapped, probably won't do that for this pass but now that we can load zaps we can start thinking about what would be best

futurepaul:
fixed the contact display and got rid of the edit button and contact activity feed because those aren't wired up yet

Screenshot 2023-09-01 at 1 04 36 PM

benthecarman:
When syncing fails there's no way to clear the error / try again

image

TonyGiorgio:
Screenshot from 2023-09-05 11-01-40

Doesn't work with this branch: https://github.com/MutinyWallet/mutiny-node/pull/742/files

.env.example
```diff
@@ -17,3 +17,4 @@ VITE_SUBSCRIPTIONS="https://subscriptions-staging.mutinywallet.com"
 VITE_STORAGE="https://storage-staging.mutinywallet.com"
 VITE_FEEDBACK="https://feedback-staging.mutinywallet.com"
 VITE_SCORER="https://scorer-staging.mutinywallet.com"
+VITE_PRIMAL="https://dev.primal.net/api"
```
TonyGiorgio, line 20: needs to be swapped to our api and needs to be added to the android build process.

src/utils/fetchZaps.ts
```diff
@@ -0,0 +1,227 @@
...
+            return undefined;
+        }
+
+        let amount = 0n;
+
+        // who is the asshole putting "lnbc9m" in all these tags?
+        if (bolt11) {
+            try {
+                // TODO: use the wallet decoder
+                // const decoded = await wallet.decode_invoice(bolt11);
```
TonyGiorgio, line 77: We can do this now.

TonyGiorgio (changes requested):

src/routes/settings/SyncNostrContacts.tsx
```diff
@@ -40,7 +40,7 @@ export function SyncContactsForm() {
     ) => {
         try {
             const npub = f.npub.trim();
-            await state.mutiny_wallet?.sync_nostr_contacts(npub);
+            await state.mutiny_wallet?.sync_nostr_contacts(undefined, npub);
```
benthecarman, line 43: this should be `VITE_PRIMAL`

benthecarman: also need to do it on SyncContactsForm.tsx

src/routes/settings/SyncNostrContacts.tsx
```diff
@@ -91,11 +94,29 @@ export function SyncContactsForm() {
...
 
     function clearNpub() {
         actions.saveNpub("");
     }
 
+    async function resync() {
+        setError(undefined);
+        setLoading(true);
+        try {
+            if (!PRIMAL_API) throw new Error("PRIMAL_API not set");
```
benthecarman, line 108: should this be outside the `try` that way it is propagated to the user and not just logged?

benthecarman (commented):
Left one comment otherwise lgtm. Just need a mutiny-node release

TonyGiorgio (approved):

src/routes/settings/SyncNostrContacts.tsx
futurepaul: it's correct that it's inside the try but what's incorrect is that I don't setError in the catch. so nice catch on my lack of catch

futurepaul: wait, no it's there github just hid it from me

TonyGiorgio:
Mainnet works pretty well for me.

PR #414, opened 2023-07-29
Translation keys 3

benalleng:
I am hoping to have 99% of translations complete by the end of this PR.

### TODO
- sanity check as to whether the key organization strategy in the translation object is a good one
- explore if we can only load the resource that the language selector chose as we currently download all languages

### WONTFIX
- due to the nature of how `<showToast />` is used in combination with resetting the entire node, this PR is unable to create a translation key for error toast titles as it would need to load the I18nContext before it has been loaded

we can delete the korean `/ko/translations.ts` before merging as ultimately they are made by a robot so there is no guarantee of their accuracy, additionally they are already a few commits old so there are already broken translations

### DIFF
--- a/src/i18n/ko/translations.ts
+++ b/src/i18n/ko/translations.ts
@@ -1,0 +1,545 @@
+export default {
+    common: {
+        title: "Mutiny 지갑",
+        nice: "멋지다",
+        home: "홈",
+        sats: "SATS",
+        sat: "SAT",
+        usd: "USD",
+        fee: "수수료",
+        send: "보내기",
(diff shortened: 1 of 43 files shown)


src/i18n/ko/translations.ts
futurepaul, line 1: I think it's fun to have a machine translation in here... whenever we announce translation stuff we might just call out that some of them are machine translated

futurepaul (approved):
epic!

futurepaul (approved):

Issue #363, opened 2023-07-15
URL bar hides receive button iphone 8


marimes:
![image](https://github.com/MutinyWallet/mutiny-web/assets/12594437/4462a10b-1a38-4317-9b3a-4874d1f88cde)

benthecarman:
Good find! If you add to home screen it will hide the navy bar

marimes:
Yes, I found that and it works 😃. However, one then needs to redo the backup as one gets new words.Am 15.07.2023 um 10:04 schrieb benthecarman ***@***.***>:﻿
Good find! If you add to home screen it will hide the navy bar

futurepaul:
this should be fixed now with the receive / send redesign #775 

<img width="487" alt="Screenshot 2024-01-13 at 1 58 22 PM" src="https://github.com/MutinyWallet/mutiny-web/assets/543668/e9825bb9-0ad9-4213-a041-7260e171c161">

PR #418, opened 2023-08-01
Design fixes: amount input scroll, fee text alignment

futurepaul:
fixing #363 was easier than I thought

while I was in there I fixed a bit of text alignment that was bothering me:

before:
Screenshot 2023-08-01 at 4 25 49 PM

after:
Screenshot 2023-08-01 at 4 31 15 PM

### DIFF
--- a/src/components/Fee.tsx
+++ b/src/components/Fee.tsx
@@ -1,0 +1,29 @@
+import { useI18n } from "~/i18n/context";
+import { AmountFiat, AmountSats } from "~/components/Amount";
+import { FeesModal } from "~/components/MoreInfoModal";
+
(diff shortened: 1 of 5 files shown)


futurepaul:
https://github.com/MutinyWallet/mutiny-web/assets/543668/ff3e412d-31d2-4cb2-beca-edeeaa983e7b

TonyGiorgio:
Screenshot from 2023-08-01 18-07-34

Seems to work well! Just wish there was some indication that a scroll is meant. Maybe an improvement in the future is to make things fit smaller, especially if we can detect if it's a small screen. Having so much text there for the warning is a bit much.

TonyGiorgio (approved):

PR #422, opened 2023-08-03
don't store default settings

futurepaul:
the old behavior was to aggressively add all environment variables to the user's localstorage. we would do this on every setup, which, for instance, overwrote a setting if the user deleted a string (like the lsp)

now we only add stuff to a new USER_ prefixed storage key when the user sets a setting to be different than the default value

this makes it possible for us to change default values for users who haven't set custom values

also I refactored a bit and the whole settings stuff is a lot more straightforward now

### DIFF
--- a/src/logic/mutinyWalletSetup.ts
+++ b/src/logic/mutinyWalletSetup.ts
@@ -3,8 +3,9 @@
 import initMutinyWallet, { MutinyWallet } from "@mutinywallet/mutiny-wasm";
 
 export type Network = "bitcoin" | "testnet" | "regtest" | "signet";
+
 export type MutinyWalletSettingStrings = {
-    network?: Network;
+    network?: string;
     proxy?: string;
     esplora?: string;
     rgs?: string;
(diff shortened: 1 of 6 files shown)


src/logic/mutinyWalletSetup.ts
```diff
@@ -15,114 +16,109 @@ export type MutinyWalletSettingStrings = {
...
+    storageKey: string,
+    defaultValue: string
+): string | undefined {
+    const item = localStorage.getItem(storageKey);
+    localStorage.ge;
+    if (item === "") {
+        return undefined;
+    } else if (item === null || item === defaultValue) {
+        return defaultValue;
+    } else {
```
benthecarman, line 77: the `|| item === defaultValue` is redundant because the `else` case will do the same

src/state/megaStore.tsx
```diff
@@ -139,11 +136,15 @@ export const Provider: ParentComponent = (props) => {
...
                 setState({ load_stage: "setup" });
 
+                const settings = await getSettings();
+
                 const mutinyWallet = await setupMutinyWallet(
                     settings,
                     password
                 );
 
+                setState({ settings: settings });
```
benthecarman, line 146: isn't this redundant, we are getting the settings right before

benthecarman (commented):
otherwise lgtm

src/state/megaStore.tsx
futurepaul: that's just setting it on the global store so other components can access settings (instead of looking it up from localstorage or re-running getsettings), I'll add a comment

TonyGiorgio (approved):

benthecarman (approved):

PR #434, opened 2023-08-07
show backup cta at launch

futurepaul:
until they back up we'll show them the notice, and now there's no dismiss button


benthecarman (approved):
After I am done backing up it takes me to the settings screen, not the home screen. Not sure if there is an easy fix for that, otherwise lgtm

PR #435, opened 2023-08-08
handle "restore" of expired subscriptions

futurepaul:
rn it fails silently if you hit restore and get back an expired timestamp. now there's an error:

Screenshot 2023-08-08 at 2 35 44 PM

### DIFF
--- a/src/utils/subscriptions.ts
+++ b/src/utils/subscriptions.ts
@@ -1,0 +1,7 @@
+const GRACE = 60 * 60 * 24 * 3; // 3 days
+
+export function subscriptionValid(subscriptionExpiresTimestamp?: number) {
+    if (!subscriptionExpiresTimestamp) return false;
+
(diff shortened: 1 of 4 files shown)


benthecarman (approved):

futurepaul:
> Did we ever have any code somewhere that gave a few days grace? I forget. I did a quick look through and didn't see it.

no right now we're pretty strict. want me to loosen it up a bit?
```
if (state.subscription_timestamp < Math.ceil(Date.now() / 1000))
```

src/utils/subscriptions.ts
```diff
@@ -0,0 +1,7 @@
+export function subscriptionValid(subscriptionExpiresTimestamp?: number) {
+    const grace = 60 * 60 * 24 * 3; // 3 days
```
benthecarman, line 2: nit could put this outside the function

TonyGiorgio (approved):

PR #440, opened 2023-08-09
Add new ui package and storybook for separation of app from components

benalleng:
Builds a scaffold for us to separate the ui from the app!

Additionally adds a storybook to allow us to test and build components more rapidly


src/components/layout/index.tsx
```diff
@@ -182,7 +183,7 @@ export const MutinyWalletGuard: ParentComponent = (props) => {
     );
 };
 
-export const LoadingSpinner = (props: { big?: boolean; wide?: boolean }) => {
+/*export const LoadingSpinner = (props: { big?: boolean; wide?: boolean }) => {
```
futurepaul, line 186: can delete this

packages/ui/src/assets/icons/airplane.svg
futurepaul, line 1: makes sense to move icons into ui but because they're copies it's a potential maintenance problem... maybe hold off for now? we don't have an example of actually using an icon from ui yeah?

futurepaul (changes requested):
getting close!

packages/ui/src/index.css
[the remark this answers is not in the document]
benalleng: Would it be impossible to change the root.css of the main project to be an import of the ui css? 

```
@import '../packages/ui/src/index.css'; 

```
Upon testing, it works.  But my gut tells me this isn't ideal

src/root.tsx
```diff
@@ -12,7 +12,8 @@ import {
     Scripts,
     Title
 } from "solid-start";
-import "./root.css";
+// eslint-disable-next-line
+import "@mutinywallet/ui/style.css";
```
benalleng, line 16: eslint hates this import, any ideas on how we can appease it?

packages/ui/postcss.config.cjs
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
futurepaul, line 1: apparently this can go into the vite? config https://github.com/MutinyWallet/mutiny-web/pull/449

src/root.tsx
futurepaul: sry :(

futurepaul (changes requested):
just need a rebase... moving the postcss stuff to vite config would be nice but if it's hassle no worries

packages/ui/postcss.config.cjs
benalleng: there is no vite config since this uses rollup to be built. maybe vite would be a better way to build the package? the postcss is used only for storybook

futurepaul (approved):
this pnpm lock makes me want to cry but I think this is progress... can perhaps revisit with a more minimal solution once the solidjs ecosystem innovates something

PR #444, opened 2023-08-14
get rid of delay for existing tab checking

futurepaul:
when I set up existing tab checking I put in 500ms timer to give it extra buffer to catch tabs... 
~~now that we have extra instance detection from the cloud I'm removing the delay~~
apparently cloud detection doesn't help this case. other two points remain:

this 1. speeds up startup times, 2. doesn't seem to miss any existing tabs in my testing anyway so maybe the buffer never helped at all

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -257,41 +257,32 @@ export const Provider: ParentComponent = (props) => {
             });
     });
 
-    // Fetch status from remote on load
     onMount(() => {
-        function handleExisting() {
-            if (state.existing_tab_detected) {
-                setState({
-                    setup_error: new Error(
-                        "Existing tab detected, aborting setup"
-                    )
-                });
-            } else {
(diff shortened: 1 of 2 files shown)


TonyGiorgio (changes requested):
The only thing with this is sometimes it may take a few seconds from opening a 2nd tab till the 2nd tab eventually times out. I see in the console logs `there's an existing tab` but it still takes a few seconds. However, this is replicated in production anyways, so it's a strict improvement. 

Though you have code quality improvements to fix. Check github actions.

futurepaul:
@TonyGiorgio alright please re-review, I ended up refactoring to simplify and clarify this logic. we had two onmounts before for no good reason, and some unecessary abstraction.

in my testing it now catches dual tabs every time and it does it before even trying to setup

TonyGiorgio (approved):
Works great!

PR #445, opened 2023-08-14
refresh sub status on load of plus page

futurepaul:
right now there's a stale renewal date on the plus page until you reload the app

### DIFF
--- a/src/routes/settings/Plus.tsx
+++ b/src/routes/settings/Plus.tsx
@@ -4,7 +4,8 @@ import {
     Suspense,
     Switch,
     createResource,
-    createSignal
+    createSignal,
+    onMount
 } from "solid-js";
 import { A } from "solid-start";
 import { ConfirmDialog } from "~/components/Dialog";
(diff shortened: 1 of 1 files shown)


TonyGiorgio (changes requested):
I subscribed, went to the home screen, refreshed, then went to the mutiny+ screen and it did not check for the subscription date again. I was monitoring the network requests.

benalleng:
I did see a get request `check-subscribed` without refreshing such that I had an accurate renewal date immediately and did not have to refresh. But I did have a separate problem, maybe unrelated/expected behavior but upon clearing my localStorage I lose my mutiny subscription. And it stopped sending the `check-subscribed` request so I could see that I had paid but was not subscribed.

TonyGiorgio:
> I did see a get request `check-subscribed` without refreshing such that I had an accurate renewal date immediately and did not have to refresh (though perhaps I am misunderstanding the problem as I see this on signet as well). But I did have a separate problem, maybe unrelated/expected behavior but upon clearing my localStorage I lose my mutiny subscription. And it stopped sending the `check-subscribed` request so I could see that I had paid but was not subscribed.

This one is kind of intentional. We don't really do phone homes for subscription checks if they're not subscribed and there's no indication of a subscription. So we have that "check subscription" button to manually check it when restores happen (which is kind of what happened in your case when you cleared local storage).

futurepaul:
gonna close until we decide what to do

PR #448, opened 2023-08-16
Move types from global.d.ts to tsconfig.json

gawlk:
Hi,

 If that's okay I'd love to help ! I'll start with some low hanging fruit issues/fixes/optimizations that I see to get to know the codebase

Anyhow, the PR is for the `vite-pwa` types in the `global.d.ts` file. Since you already use the `compilerOptions.types` attribute in the `tsconfig` (and they basically do the same thing) I figured one less file at the root of the project would be better

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,1 +1,0 @@
-/// <reference types="vite-plugin-pwa/client" />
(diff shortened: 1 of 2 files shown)


futurepaul (approved):
hey, welcome! I love little cleanups like this, it's much appreciated!

gawlk:
You're very welcome, I hope I'll be able to help some more

PR #449, opened 2023-08-16
Move postcss config to vite's config file

gawlk:
Just another simple change, moving the config to `vite.config.ts` allows one less file and having types

They aren't used right now and the config will most likely not be needed after the integration of the lightning css engine but still


futurepaul (approved):
another great win, thank you

PR #450, opened 2023-08-16
Prettier improvements

gawlk:
- Updated `Prettier` to 3.0.2
- Deleted the `prettier-ignore` file since [it can read the `.gitignore` file now](https://github.com/prettier/prettier/pull/14731)
- Added the `prettier-tailwind-plugin`, [the repo will probably explain the pros better than me](https://github.com/tailwindlabs/prettier-plugin-tailwindcss) but it's basically for consistency and a visibility of classes priorities over each other
- Didn't run prettier over the codebase for a cleaner PR


gawlk:
I'm not sure what's up with the lock file, it looks fine on my end 🤔

pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
benalleng, line 1: I can guarantee this is your problem, I have the same issue regularly.

gawlk: It worked ! Thanks, I wonder what's the underlying issue

gawlk:
The warnings are expected since I didn't run prettier on purpose

pnpm-lock.yaml
benalleng: who knows :shrug:

package.json
```diff
@@ -11,13 +11,14 @@
...
-        "format": "npx prettier --write \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
-        "check-format": "npx prettier --check \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\""
+        "format": "prettier --write \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
+        "check-format": "prettier --check \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\""
     },
     "type": "module",
     "devDependencies": {
         "@capacitor/assets": "^2.0.4",
         "@capacitor/cli": "^5.2.2",
+        "@ianvs/prettier-plugin-sort-imports": "^4.1.0",
```
futurepaul, line 21: this seems great but as a security concern I try to pretty conservative with adding dependencies, same with prettier-plugin-tailwind

would much prefer if these could just be binary tools instead of dependencies!

.prettierrc
```diff
@@ -1,9 +1,18 @@
...
+    "arrowParens": "always",
+    "printWidth": 80,
+    "useTabs": false,
+
+    "plugins": [
+        "prettier-plugin-tailwindcss",
+        "@ianvs/prettier-plugin-sort-imports"
+    ],
+
+    "tailwindFunctions": ["cva"],
```
futurepaul, line 15: nice addition but we'll be getting rid of cva in #459

futurepaul (changes requested):
more helpful cleanup thank you. I'd like to avoid more dependencies for now, but I might run prettier-plugin-tailwindcss and @ianvs/prettier-plugin-sort-imports as a one-off cleanup just to get stuff in a nicer state, so thanks for the suggestion.

package.json
gawlk: I understand (but `prettier-plugin-tailwind` is made by tailwind labs, if this dependency can be compromised `tailwindcss` can be too)

Am I right to assume that it is to protect your working environment (and the team's) and not mine for example, or the cloud flare page build step, etc ?

If so I might have a solution with optional (well peer) deps, checks if they're installed in prettier (or create a wrapper config if too hard) and maybe to get fancy a GitHub action with weekly PRs so that you don't even have to think about it.

.prettierrc
gawlk: I saw that but thought it could be here until the rewrite is done, but I did forget `classList`

package.json
gawlk: No matter what I do (optional, peer, optional peer) `pnpm` installs all reps by default.. I can still do the action though

gawlk:
Alright so I got the action to work on a copy if this repo (https://github.com/gawlk/mutiny-with-prettier-workflow/actions) without any token but I did need to allow create and *approvals* of PR which is supposed isn't great for you.
But you could use your own token for the PR so it shouldn't be an issue. I can try to see how it works if you want

So basically I made an optional config inside the prettier config file which is added only if the dependencies are installed. The workflow pull the repo and adds all unsafe deps and runs prettier, then it creates a branch and a PR which you.

The workflow runs once a week with the possibility to run it manually if needed

gawlk:
And if neither possibilities are secure enough for you

I see two more solutions:
- Remove the workflow while keeping the rest, and move it to a synced fork of the repo which would do the PR automatically every week or so, which would be like any other contribution
- Run a workflow in every PR but auto commit would probably be annoying

gawlk:
Hey @futurepaul, what would you like me do about the lock file ? Doing it manually would be a horror and there is an update so the file has to change

futurepaul:
Appreciate all this effort to make this work. I think you're right about `prettier-plugin-tailwindcss` that can be added as a dev dep because it's from the same provenance...

The more that I think about this, the more comfortable I am with your original incarnation. Any diff will be evident in PRs, so as long as I keep reviewing PRs we should be good. Sorry for the run around. I am stoked to think of the possibilities of a workflow that creates prs, but for now I'd rather not have all that noise :)

some clarity about the pnpm lockfile, hopefully a rebase on top of this will work? (make sure you've updated pnpm): https://github.com/MutinyWallet/mutiny-web/pull/463

futurepaul:
just to be clear: please change this back to how it was originally, with both the import sort and tailwind prettier added to the format command

gawlk:
No worries, I fully understand that security for such an app is extremely important

gawlk:
Squashed and rebased should be good to go (but still didn't run `pnpm format`, should I ?)

pnpm-lock.yaml
benalleng: Fixed by #463

futurepaul (approved):
appreciate this, long overdue cleanup, but hopefully this is the last in your series of diffsplosions!

gawlk:
Ready to roll 🤙

gawlk:
> appreciate this, long overdue cleanup, but hopefully this is the last in your series of diffsplosions!

Please don't hate me but I still have quite a few cleanups in mind 🥹

PR #451, opened 2023-08-16
Fix tagToMutinyTag ts error

gawlk:
Fixed the TS error, not sure if you're fans or haters of `switch` though, of course will change to ifs if preferred.

But one weird thing is that the type of `kind` is supposed to be a number in the input and a string in the output. Either the input is badly typed or there might be some bugs in the app you use `kind` somewhere since the return type was lying to you until now

I'll edit this PR after I check which one it is 

EDIT: Kind of overwhelmed for now by the codebase, I'll need some time to figure how to pass through this function and print a `console.log` (to know what's in the original `tag.kind`)

### DIFF
--- a/src/utils/tags.ts
+++ b/src/utils/tags.ts
@@ -25,8 +25,21 @@ export function tagsToIds(tags?: MutinyTagItem[]): string[] {
 }
 
 export function tagToMutinyTag(tag: TagItem): MutinyTagItem {
-    // @ts-expect-error: FIXME: make typescript less mad about this
-    return tag as MutinyTagItem;
+    let kind: MutinyTagItem["kind"];
+
+    switch (tag.kind) {
+        case 0: {
+            kind = "Label";
+            break;
+        }
+        case 1:
+        default: {
+            kind = "Contact";
+            break;
+        }
+    }
+
+    return { ...tag, kind };
 }
 
 export function sortByLastUsed(a: MutinyTagItem, b: MutinyTagItem) {


futurepaul:
yeah tbh our typescript story is a little hacky. sometimes we get really good types from the wasm-bindgen, but some rust types have to be exported as a js `any` so I create these dummy types for them. long term it would be nice to have the source of truth for types like this in the `mutiny-wasm` project, or at least one spot inside the `mutiny-web` project

oh IRT console.log: for some complex structs we actually get a pointer, not a normal js object, so you can't print them. we try to add a `.value()` method to types so they can be printable, but haven't been super consistent I'm sure

I've also noticed weirdness with an enum type from rust "lying" about what it is

here's the mutiny-wasm code for reference:
https://github.com/MutinyWallet/mutiny-node/blob/master/mutiny-wasm/src/models.rs#L632

my guess in this case the hackiness resulted because I want to switch on a string, but under the hood the `TagKind` enum is encoded as a number

gawlk:
Thank you for the insight, very informative !

I actually saw the enum at the very top of: /node_modules/.pnpm/@mutinywallet+mutiny-wasm@0.4.12/node_modules/@mutinywallet/mutiny-wasm/mutiny_wasm.d.ts

So the types aren't wrong for this, it's just the testing if it's really a number that I couldn't do

futurepaul (approved):
thx sry wanted to soak on this one for a bit but it looks like it gets the job done

PR #456, opened 2023-08-16
add safe mode view

futurepaul:
if we're having connection problems there's a link to a safe mode `/?safe_mode=true` that disables lightning


benthecarman (commented):
need to update to 0.4.12 in here

benthecarman (approved):

PR #457, opened 2023-08-16
Add .env to ignored files

gawlk:
This app needs it the function in dev mode and yet it's not in the repo so I suppose you just forgot to add it to the gitignore file ?


futurepaul (approved):
yeah I haven't needed because we deploy to cloudflare and I use `.env.local` for my local testing, but it can't hurt!

PR #459, opened 2023-08-16
Rewrite button.tsx to remove class-variance

### LINKED ISSUE #413: Rewrite `Button` component
I think it's served us long enough but we can get rid of the `class-variance` dependency by just using solid's built-in `classList` stuff

benalleng:
Closes #413 

Rewrites `<Button />` and `<ButtonLink />` to no longer use cva and instead uses classList for most cases. ~~@futurepaul and I have had discussion about avoiding string interpolation in `class={'${props.class}'}` but alas I did not think of a great way to do that here as I found several instances of the button components where they used individual classes.~~

also adds a small fix for a prop bug in the `<Collapser />` component

### DIFF
--- a/src/components/ContactForm.tsx
+++ b/src/components/ContactForm.tsx
@@ -49,9 +49,11 @@ export function ContactForm(props: {
                     </Field> */}
                 </VStack>
             </div>
-            <Button type="submit" intent="blue" class="w-full flex-none">
-                {props.cta}
-            </Button>
+            <VStack>
+                <Button type="submit" intent="blue">
+                    {props.cta}
+                </Button>
+            </VStack>
         </Form>
     );
 }
--- a/src/components/layout/Button.tsx
+++ b/src/components/layout/Button.tsx
@@ -1,69 +1,57 @@
 import { LoadingSpinner } from "@mutinywallet/ui";
-import { cva, VariantProps } from "class-variance-authority";
-import { children, JSX, ParentComponent, Show, splitProps } from "solid-js";
+import { JSX, ParentComponent, Show, splitProps } from "solid-js";
 import { Dynamic } from "solid-js/web";
 import { A } from "solid-start";
 
-const button = cva(
-    "p-3 rounded-xl font-semibold transition disabled:bg-neutral-400/10 disabled:text-white/20 disabled:shadow-inner-button-disabled",
-    {
-        variants: {
-            // TODO: button hover has to work different than buttonlinks (like disabled state)
-            intent: {
-                active: "bg-white text-black border border-white hover:text-[#3B6CCC]",
-                inactive:
-                    "bg-transparent text-white border border-white hover:text-[#3B6CCC]",
-                glowy: "bg-black/10 shadow-xl text-white border border-m-blue hover:m-blue-dark hover:text-m-blue",
-                blue: "bg-m-blue text-white shadow-inner-button hover:bg-m-blue-dark text-shadow-button",
-                red: "bg-m-red text-white shadow-inner-button hover:bg-m-red-dark text-shadow-button",
-                green: "bg-m-green text-white shadow-inner-button hover:bg-m-green-dark text-shadow-button",
-                text: ""
-            },
-            layout: {
-                flex: "flex-1 text-xl",
-                pad: "px-8 text-xl",
-                small: "px-4 py-2 w-auto text-lg",
-                xs: "px-4 py-2 w-auto rounded-lg text-base",
-                full: "w-full text-xl"
-            }
-        },
-        defaultVariants: {
-            intent: "inactive",
-            layout: "flex"
-        }
-    }
-);
-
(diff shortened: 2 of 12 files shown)


src/components/ContactForm.tsx
```diff
@@ -44,7 +44,7 @@ export function ContactForm(props: {
                     </Field> */}
                 </VStack>
             </div>
-            <Button type="submit" intent="blue" class="w-full flex-none">
+            <Button type="submit" intent="blue">
```
benalleng, line 47: "noflex"

src/components/layout/Button.tsx
```diff
@@ -1,69 +1,62 @@
...
-                layout: local.layout
-            })}
+            class="p-3 rounded-xl font-semibold transition disabled:bg-neutral-400/10 disabled:text-white/20 disabled:shadow-inner-button-disabled"
+            classList={{
+                "bg-white text-black border border-white hover:text-[#3B6CCC]":
+                    props.intent === "active",
+                "bg-transparent text-white border border-white hover:text-[#3B6CCC]":
+                    !props.intent || props.intent === "inactive",
+                "bg-black/10 shadow-xl text-white border border-m-blue hover:m-blue-dark hover:text-m-blue":
+                    props.intent === "glowy",
```
futurepaul, line 44: we can get rid of "glowy" for this and button link, it's only used in KitchenSink fine to delete from its uses there and just have default button

src/components/layout/Button.tsx
```diff
@@ -1,69 +1,62 @@
...
+                "bg-m-blue text-white shadow-inner-button hover:bg-m-blue-dark text-shadow-button":
+                    props.intent === "blue",
+                "bg-m-red text-white shadow-inner-button hover:bg-m-red-dark text-shadow-button":
+                    props.intent === "red",
```
futurepaul, line 48: these intents have a lot of duplicate stuff:
```text-white shadow-inner-button text-shadow-button```

File: src/components/layout/Button.tsx [hunk 4]
```diff
@@ -1,69 +1,62 @@
...
+    href: string;
+    target?: string;
+    rel?: string;
+};
 
+export const Button: ParentComponent<ButtonProps> = (props) => {
     return (
         <button
-            {...attrs}
+            onClick={props.onClick}
```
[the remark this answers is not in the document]
benalleng: unfrtunately, when I do that all buttons stop working

src/components/layout/Button.tsx
benalleng: want to try intent="red" for those?

src/components/layout/Button.tsx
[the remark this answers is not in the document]
benalleng: *it does get rid of the wanring though

gawlk:
Just in case, I thought I'd share how I ended up fixing this issue in my projects/day job

Instead of using `classList` which is nice but has everything very flat and can end up quite confusing, I put everything in class in created a function that takes an input (array of string/array/false/undefined, or a string directly) and builds a clean string from it, it also makes the `class` attribute nicer to use and much more flexible

Here's one component: https://gitlab.com/isaan/mapview-dev/mapview2/-/blob/solid/src/components/generic/container/component.tsx?ref_type=heads#L25

src/components/layout/Button.tsx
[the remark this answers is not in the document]
futurepaul: lol

futurepaul:
my superstition against runtime string building stems from tailwind skipping over values that it doesn't find during "build", which has bit me in the ass at the most surprising times because tailwind might have a style cached so I see it during dev but during actual build it's not there... obviously there's a way that libs like CVA solve this but I haven't quite figured it out and don't see how your method fixes that either

futurepaul:
@benalleng still trying to figure out types for onClick? ideally this can just extend normal button types shouldn't have to include onclick explicitly in your prop def... that was the purpose of the
```
interface ButtonProps
    extends JSX.ButtonHTMLAttributes<HTMLButtonElement>,
```
stuff in there before

gawlk:
> my superstition against runtime string building stems from tailwind skipping over values that it doesn't find during "build", which has bit me in the ass at the most surprising times because tailwind might have a style cached so I see it during dev but during actual build it's not there... obviously there's a way that libs like CVA solve this but I haven't quite figured it out and don't see how your method fixes that either

I've been using tailwind for a few years now and never encountered similar problems so I couldn't tell. If I understand the problem correctly, I avoid not found classes by following the tailwind's guidelines which in this case are classes are in a class attribute, and full (using something like `text-red-{value}` is forbidden)

Fixing wasn't the right word for sure, I just meant to send my way of solving the variance problem

futurepaul (approved):
works / looks great, thank you! 

@gawlk took me a while to really understand your approach. it's pretty cool and really readable (once I got over those `run` functions) but I'm happy with classlist for now

PR #463, opened 2023-08-17
fresh pnpm install just to make sure

futurepaul:
I ran a fresh pnpm install so I would know we have a lockfile in a known good case

from looking at this issue, 6.1 was a mistake, 6.0 is the correct version:
https://github.com/pnpm/pnpm/issues/6648#issuecomment-1588863457

@benalleng @gawlk make sure you guys aren't using one of these bad pnpm versions: v8.6.0, v8.6.1, v7.33.0, that seems to be what introduces the 6.1

I updated the package.json to match cloudflare's installed version of pnpm (which is the main thing we need to keep compatiblity with): https://developers.cloudflare.com/pages/platform/language-support-and-tools/

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "mutiny-wallet",
     "version": "0.4.12",
     "license": "MIT",
-    "packageManager": "pnpm@8.3.1",
+    "packageManager": "pnpm@8.6.6",
     "scripts": {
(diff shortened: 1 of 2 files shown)


benalleng (approved):
Looks like the bug is fixed as of 8.6.2 so this should be good!
Reading that pnpm issue gave me an aneurysm

PR #464, opened 2023-08-17
approve all

futurepaul:
https://github.com/MutinyWallet/mutiny-web/assets/543668/5b0aaf98-f1c0-4d51-8b17-6a533f20e915

### DIFF
--- a/src/components/PendingNwc.tsx
+++ b/src/components/PendingNwc.tsx
@@ -79,6 +79,14 @@ export function PendingNwc() {
         }
     }
 
+    async function approveAll() {
+        // clone the list so it doesn't update in place
+        const toApprove = [...pendingRequests()!];
+        for (const item of toApprove) {
+            await payItem(item);
+        }
+    }
+
     async function rejectItem(item: PendingItem) {
         try {
             setPaying(item.id);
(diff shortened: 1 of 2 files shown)


src/components/PendingNwc.tsx
```diff
@@ -181,12 +189,22 @@ export function PendingNwc() {
...
-                    class="text-m-red active:text-m-red/80 font-semibold no-underline self-center"
-                >
-                    {i18n.t("settings.connections.pending_nwc.configure_link")}
-                </A>
+                <div class="flex justify-around w-full">
+                    <button
+                        class="text-m-green active:text-m-red/80 font-semibold"
+                        onClick={approveAll}
+                    >
+                        Approve All
```
benthecarman, line 197: should make this translatable

src/components/PendingNwc.tsx
```diff
@@ -81,6 +81,14 @@ export function PendingNwc() {
         }
     }
 
+    async function approveAll() {
+        // clone the list so it doesn't update in place
+        const toApprove = [...pendingRequests()!];
+        for (const item of toApprove) {
+            await payItem(item);
```
benthecarman, line 88: we should handle if this fails, at the very least log it. Adding specific errors to the item would be ideal but that might be harder

futurepaul: each "payitem" call handles its own errors

benthecarman (approved):

PR #465, opened 2023-08-17
use local version of spinner not ui lib

futurepaul:
some kinks to work out in ui library, for now I'm going to go back to the local version of the loading spinner while we try to figure out a nicer solution


benalleng (approved):
We'll figure it out! Sorry I didn't get it quite right. Looks good.

futurepaul:
closing in favor of #467

PR #467, opened 2023-08-18
Fix ui lib

benalleng:
Tailwind config was missing the ui/src content so it was being lost when compiling.

Additionally, I had some trouble consistently building the ui package so I am changing the `build-submodules` script.

### TODO
- looking into adding a watcher / input on the ui so that file changes there will restart the dev server ultimately removing the need for constant rebuilds

[Storybook Preview](https://ff5ed1f2.mutiny-web.pages.dev/storybook)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
     "license": "MIT",
     "packageManager": "pnpm@8.6.6",
     "scripts": {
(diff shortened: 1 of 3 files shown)


futurepaul (approved):
works on my machine!

PR #469, opened 2023-08-18
Add nwc callback query parameters

TonyGiorgio:
can test with this url: 

https://nwc-callback.mutiny-web.pages.dev/settings/connections?callbackUri=primal&name=Primal

### DIFF
--- a/src/i18n/ko/translations.ts
+++ b/src/i18n/ko/translations.ts
@@ -253,9 +253,9 @@ export default {
             new_connection_label: "이름",
             new_connection_placeholder: "내가 좋아하는 nostr 클라이언트...",
             create_connection: "연결 생성",
-            open_app: "Open App",
-            open_in_nostr_client: "Open in Nostr Client",
-            open_in_primal: "Open in Primal",
+            open_app: "앱 열기",
+            open_in_nostr_client: "Nostr 클라이언트에서 열기",
+            open_in_primal: "Primal에서 열기",
             authorize:
                 "외부 서비스가 지갑에서 결제를 요청할 수 있도록 인증합니다. nostr 클라이언트와 잘 맞습니다.",
             pending_nwc: {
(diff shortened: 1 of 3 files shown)


benthecarman (approved):

TonyGiorgio:
Ready for review @futurepaul 

@benalleng if you have any suggestions for the korean words, would appreciate a quick look.

src/i18n/ko/translations.ts
```diff
@@ -253,6 +253,9 @@ export default {
             new_connection_label: "이름",
             new_connection_placeholder: "내가 좋아하는 nostr 클라이언트...",
             create_connection: "연결 생성",
+            open_app: "Open App",
+            open_in_nostr_client: "Open in Nostr Client",
+            open_in_primal: "Open in Primal",
```
benalleng, line 258: To be clear 99% of the translations are done unreviewed by chat gpt, but these are pretty easy.

```suggestion
            open_app: "앱 열기",
            open_in_nostr_client: "Nostr 클라이언트에서 열기",
            open_in_primal: "Primal에서 열기"
```

And below is the full "Korean-ification"

```suggestion
            open_app: "앱 열기",
            open_in_nostr_client: "노스트 클라이언트에서 열기",
            open_in_primal: "프라이멀에서 열기"
```

benalleng: As a caveat, open app is the literal translation, If you mean "Open in the app" you might want to use "앱에서 열기"

futurepaul (approved):
this works for me. only caveat is that the params stay in place after use so if someone creates another one without doing other navs inside mutiny first they get re-prompted to open the app... not a big deal I can handle in a followup if it bugs me too much

src/i18n/ko/translations.ts
TonyGiorgio: There's two suggestions above, which are you saying is the one that is fine?

benalleng: Sorry, let's go with the first one, the second was just turning proper nouns, nostr and Primal into Korean letters

PR #472, opened 2023-08-20
Clear error after send button disabled and enhance error catches

### LINKED ISSUE #462: Changing send amount after too high requires starting over
1. Have an amountless invoice or keysend to a node
2. pick an amount over your balance
3. get the error that amount is too high
4. change the amount to 1 sat (singular, not plural....)
5. the error still doesn't let you send

signal-2023-08-17-134116_002

benalleng:
Closes #462 

Also included a very obscure catch wherby the send button is disabled if there is and amount in the invoice but it does not equal amountSats() `invoice().amount_sats !== amountSats()` unsure if/how this would ever occur as having the ability to edit the amount is only possible if there is no amount set in the invoice. If it seems unneccessary I'm happy to remove it

This video below is only possible because of some altered logic `isAmountEditable={true}`

[Kazam_screencast_00003.webm](https://github.com/MutinyWallet/mutiny-web/assets/108441023/ee8ce4c9-e24b-4e55-84c5-4fbff41bd5a0)

### DIFF
--- a/src/i18n/en/translations.ts
+++ b/src/i18n/en/translations.ts
@@ -109,6 +109,8 @@ export default {
         },
         error_low_balance:
             "We do not have enough balance to pay the given amount.",
+        error_invoice_match:
+            "Amount requested, {{amount}} SATS, does not equal amount set.",
         error_clipboard: "Clipboard not supported",
         error_keysend: "Keysend failed",
         error_LNURL: "LNURL Pay failed"
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -246,6 +246,15 @@ export default function Send() {
         setFieldDestination("");
     }
 
+    // If we got here from a scan result we want to set the destination and clean up that scan result
+    onMount(() => {
+        if (state.scan_result) {
+            setDestination(state.scan_result);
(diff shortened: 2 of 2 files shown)


futurepaul:
@benalleng I wanted to clean up some of this logic and figured it would be easiest to just commit on top of your pr because I'm cleaning up my own mess.

basically it was working to run side effects during the recomputation of a reactive value like `const insufficientFunds = () =>` but it's confusing / hard to reason about. it meant that multiple "reactive" things were fighting over the state of `error()`.  now it should be clearer which of these are effectful functions vs reactive values.

would appreciate you checking this over to make sure we're still catching all the same errors we want to be. it was working for me with on-chain and amountless invoice so I feel pretty good about it

benalleng:
@futurepaul tack! Went through and it resets properly. reached each possible error and then cleared upon changing the send amount. using a createEffect makes a lot more sense! Good call.

How did you make an amountless invoice for signet?

benthecarman:
@benalleng you can do this to get an amountless invoice

```
curl -X POST https://faucet.mutinynet.com/api/bolt11 -H 'Content-Type: application/json' -d '{}'
```

PR #474, opened 2023-08-21
Delete all for amountEditable should use PointerEvents

benalleng:
Fixes a bug where the hold "DEL" to delete all only worked on Desktop, `PointerEvents` should work on all devices. Additionally did not replace `onMouseLeave` with `onPointerLeave` as it isn't necessary in this logic


futurepaul (approved):
works great nice fix

PR #481, opened 2023-08-22
Prod push

TonyGiorgio:
Just the nwc link stuff. I don't see anything else immediate down the pipeline so let's get this out.


futurepaul (approved):
first update in a while without a mutiny-web version bump. shouldn't be a problem just worth noting in case caching gets weird

TonyGiorgio:
> first update in a while without a mutiny-web version bump. shouldn't be a problem just worth noting in case caching gets weird

we've done them before without a problem.

PR #483, opened 2023-08-22
Handle back button navigation on android

TonyGiorgio:


futurepaul (approved):
works on my phone!

ideally we'd have a "pop stack" nav, rather than using window history because:

if you go to settings, then click the back link in the app, then click the back button it goes back to settings...

so anyway something to keep in mind I don't think this is exact native behavior but def a big improvement

TonyGiorgio:
> works on my phone!
> 
> ideally we'd have a "pop stack" nav, rather than using window history because:
> 
> if you go to settings, then click the back link in the app, then click the back button it goes back to settings...
> 
> so anyway something to keep in mind I don't think this is exact native behavior but def a big improvement

agreed, i noticed that too. Or at least pulling in the `BackLink` functionality to mimic that. But at least it doesn't exit the app which is what happens now and is pretty jarring. Worst case the user corrects it themselves to the right screen.

PR #486, opened 2023-08-23
Update to 0.4.13

benthecarman:
looks like `pnpm install` updated a couple other deps, not sure if that is okay

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "com.mutinywallet.mutinywallet"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 10
-        versionName "0.4.12-2"
+        versionCode 11
+        versionName "0.4.13-1"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
              // Files and dirs to omit from the packaged assets dir, modified to accommodate modern web apps.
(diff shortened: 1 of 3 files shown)


TonyGiorgio:
are there any breaking API calls? If so it might be best to include this in the gifting PR.

benthecarman:
> are there any breaking API calls? If so it might be best to include this in the gifting PR.

looking through the diff, there shouldn't be

benthecarman:
tests failing, idk why, maybe i'll just let @futurepaul handle it lol

benalleng:
seems we have broken storage

image

the test fails after encrypting

futurepaul:
yeah good find @benalleng

@benthecarman pretty easy to repro:
go to backup, check the three boxes, then enter an encryption password and hit save. you get that `recursive use` error and then on reload storage is broken

futurepaul:
hmm might not be that cut and dry

if you go to https://signet-app.mutinywallet.com/settings/encrypt and click "encrypt" without entering a password you get the same recursive error, there's just no problem on load. so it's something to do with the decrypt step not the encrypt step I think

benthecarman:
I can reproduce, but everything seems to work? Not sure what the best course of action is here

futurepaul:
New theory: you're returning the "Failed to read data from storage" error rather than the "Incorrect password entered". I'm watching for the incorrect password error so I know to prompt the user to inter their password. if I just hardcode it to prompt the user to enter their password it works fine

TonyGiorgio:
> if you go to https://signet-app.mutinywallet.com/settings/encrypt and click "encrypt" without entering a password you get the same recursive error, there's just no problem on load. so it's something to do with the decrypt step not the encrypt step I think

Nothing's changed here, it's a flacky test.

PR #491, opened 2023-08-25
Backup seedwords UX

benalleng:
On testing the last version upgrade, I noticed the backup button was lacking in styling as compared to a standard button. 
This is mostly a styling change but should allow for some better UX with the hide show being used


futurepaul (changes requested):
I appreciate the attempt here but don't think it's really correct to turn the whole red box into a a "button" stylistically. I'll think about it some more maybe do a figma. for now I don't think this works

benalleng:
Closing to wait for an official style decision

PR #492, opened 2023-08-25
Encryption password UX

### LINKED ISSUE #490: Disallow same password when changing password
Right now if you change your encryption password to the same thing it will allow it. We should probably forbid this for simplicity sake

benalleng:
Closes #490 after [Mutiny-node/#727](https://github.com/MutinyWallet/mutiny-node/pull/727) is merged

Adds similar logic to prevent #462 from cropping up here

### DIFF
--- a/src/routes/settings/Encrypt.tsx
+++ b/src/routes/settings/Encrypt.tsx
@@ -1,5 +1,5 @@
 import { createForm } from "@modular-forms/solid";
-import { createSignal, Show } from "solid-js";
+import { createMemo, createSignal, Show } from "solid-js";
 
 import {
     BackLink,
@@ -31,7 +31,7 @@ export default function Encrypt() {
     const [error, setError] = createSignal<Error>();
     const [loading, setLoading] = createSignal(false);
 
-    const [_encryptPasswordForm, { Form, Field }] =
+    const [encryptPasswordForm, { Form, Field }] =
         createForm<EncryptPasswordForm>({
             initialValues: {
                 existingPassword: "",
@@ -39,10 +39,15 @@ export default function Encrypt() {
                 confirmPassword: ""
             },
             validate: (values) => {
+                setError(undefined);
                 const errors: Record<string, string> = {};
                 if (values.password !== values.confirmPassword) {
                     errors.confirmPassword = i18n.t(
                         "settings.encrypt.error_match"
+                    );
+                } else if (values.password === values.existingPassword) {
+                    errors.password = i18n.t(
+                        "settings.encrypt.error_same_as_existingpassword"
                     );
                 }
                 return errors;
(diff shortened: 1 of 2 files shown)


benalleng:
drafting now that it has been pointed out to me a createEffect on password change might be a more elegant way to use this logic

src/routes/settings/Encrypt.tsx
```diff
@@ -66,6 +67,14 @@ export default function Encrypt() {
         }
     };
 
+    const encryptButtonDisabled = createMemo(() => {
+        return (
+            !_encryptPasswordForm.touched ||
```
futurepaul, line 72: the `_` prefix is so I can keep a variable around unused and keep the linter happy, can remove the underscore

src/routes/settings/Encrypt.tsx
```diff
@@ -39,6 +39,7 @@ export default function Encrypt() {
                 confirmPassword: ""
             },
             validate: (values) => {
+                setError();
                 const errors: Record<string, string> = {};
                 if (values.password !== values.confirmPassword) {
```
benthecarman, line 44: could also add the check here that old != new password

benalleng: I had that initially but with your mutiny-node pr I thought it might be error code overload, what do you think?

benalleng: I had that initally but I though that might be error code overload, what do you think?

benthecarman: May as well defend on both ends 🤷‍♂️

src/routes/settings/Encrypt.tsx
benalleng: didn't know that! tmyk :brain:

src/routes/settings/Encrypt.tsx
benalleng: nvm I see what you mean, validate runs before handleEncrypt so no double errors

benalleng:
Upon learning how modular forms works more, I don't createEffect is a good route afterall, the standard validate after submitting is sufficient. It then continuosly checks on its own after submitting already.

src/routes/settings/Encrypt.tsx
```diff
@@ -31,19 +31,24 @@ export default function Encrypt() {
...
-    const [_encryptPasswordForm, { Form, Field }] =
+    const [encryptPasswordForm, { Form, Field }] =
         createForm<EncryptPasswordForm>({
             initialValues: {
                 existingPassword: "",
                 password: "",
                 confirmPassword: ""
             },
             validate: (values) => {
+                setError(undefined);
```
futurepaul, line 42: makes more sense to put this in `handleFormSubmit` because that's where the setter happens

futurepaul (changes requested):
looks / works great just needs rebase. thank you!

src/routes/settings/Encrypt.tsx
benalleng: Just tried this, but I remembered why its here, we need the error to reset on any change to the form and not only on submit since an error prevents the user from submitting.

benalleng: Due to the `encryptButtonDisabled` the encrypt button is unclickable until the user has made a change to any of the forms

futurepaul: ahh k makes sense

futurepaul (approved):
<3

PR #494, opened 2023-08-28
Sorted contacts on activtiy screen

benthecarman:
This makes it so the activity screen will have the top contacts always be the same order (sorted by last used then by name). `get_contacts` was a json object so the ordering would be pretty much random. Now we will get it in a sorted array


futurepaul (approved):
nice

PR #498, opened 2023-08-29
Prod 0.4.14

TonyGiorgio:


futurepaul (approved):
lgtm sorry about this restore test I think it has something to do with timing of loading the "same" wallet

benthecarman (approved):

PR #500, opened 2023-08-29
Autoverify open by default links

benalleng:
Should allow the app to have app.mutinywallet.com be autoverified

turns out the data actually doesn't need to be in separate intents

### TODO
make sure we are using the correct fingerprint for our release in assetlinks.json


TonyGiorgio:
oh yeah good catch, I forgot the autoverify part from when i was testing it

TonyGiorgio (approved):

PR #501, opened 2023-08-29
Budget NWC Testing

benthecarman:
this is just how I am testing the budget nwcs


futurepaul:
Screenshot 2023-09-12 at 3 59 53 PM

got this editor wired up seems to work pretty well. don't have a "save" step, just update the profile as they change the amount / pick the day / toggle auto approve

src/routes/settings/Connections.tsx
```diff
@@ -25,13 +35,185 @@ import {
...
+        } catch (e) {
+            console.error(e);
+        }
+    }
+
+    async function saveNewBudget(amountSats: bigint) {
+        try {
+            await state.mutiny_wallet?.set_nwc_profile_budget(
+                props.profile.index,
+                amountSats,
```
benalleng, line 109: Not to needlessly complicate the code but since the budget is implied as a total remaining should we set the budget amount as:
`amountSats - (props.profile.budget_amount - props.profile.budget_remaining)`
By setting it to just amountSats any change will completely reset the budget.

benalleng: I think..? this would also fix the issue where changing the day also resets the budget... not 100% on this one

benalleng (changes requested):
Works great for me! Still think there are a few ways to enchance it namely not reseting the remaining amount unless the user absolutely wants that.

src/routes/settings/Connections.tsx
benalleng: I misunderstood how the set_nwc_profile_function worked in my previous comment

src/routes/settings/Connections.tsx
```diff
@@ -25,13 +35,185 @@ import {
...
+                0
```
benthecarman, line 110: This is setting the `BudgetPeriod` to 0, not sure what that will do.

src/routes/settings/Connections.tsx
```diff
@@ -25,13 +35,185 @@ import {
...
+        Browser.open({ url: connectString });
+    }
+
+    async function setAutoApprove(checked: boolean) {
+        try {
+            if (checked) {
+                await state.mutiny_wallet?.set_nwc_profile_budget(
+                    props.profile.index,
+                    420n,
+                    0
```
benthecarman, line 92: same here

TonyGiorgio (changes requested):
1. It should not create by default a budget nwc
2. ~~It should at minimum automatically open the nwc details so they may be configured~~ Actually I take this back, it won't play well with the auto redirect. The creation form with the name should probably have the same editable information.
3. It should have some sort of update button
4. Editing an NWC is resetting the budget. Is that intended? Part of the reason why we need a save button. We should prompt them that editing an NWC budget when there's already a history is going to reset the budget for the duration specified. 
5. When turning on auto approval we should probably display a confirmation that it means that someone that gets access to this string can automatically pull funds from the wallet up to the budget in that period, and to safeguard it.

futurepaul:
okay thx for all the feedback was an oversight that I left in the predefined budget stuff

I moved all the budgeting to the creation modal
Screenshot 2023-09-14 at 4 49 50 PM

I could ALSO have budget editing and I know we have all the methods for it... but how often is that really needed? feels like it clutters up the ui and it's not insanely hard to delete and recreate if you need to? especially with the confusion around what happens to your existing payments in a current budget

benalleng:
> how often is that really needed?

I would guess there is really only one time editing is needed; when the user runs out of money before the end of the budget period and wants to expand / top-up the budget.

Going the other direction, i.e. realizing that they have a budget for Primal = 1BTC / second and they want to reduce they would likely just create a new budget profile.

futurepaul:
> wants to expand / top-up the budget permanently or until the end of this period.

a "refill" function might work for that. otherwise how is it obvious that "editing" the budget resets it for this period?

benalleng:
I don't think it is, maybe we are talking past each other a little bit for this one but I think that my primary concern is centered around the longer periods if they edit it in the middle of the year completely reseting it.  The shorter the period the less it matters.

Ultimately since its primarily zaps its all being controlled by the user. No disney+ auto-withdraws... yet lol

benthecarman (commented):
ACK

benalleng (changes requested):
Feels great! 
One small thing; I am allowed to set my budget below what my remaining budget resulting in this error.
image

src/components/NWCBudgetEditor.tsx
benalleng, line 1: Think it would be feasible to bring the budget.remaining state into this file to compare and present and error to the user?

benthecarman: Fixed this on the backend side https://github.com/MutinyWallet/mutiny-node/pull/761

we could cut another release or just do some sanity checks here to prevent the massive numbers like that

benalleng: new node version is tACK from me, up to you guys on whether you think its worth having an error in the frontend for that user-flow

TonyGiorgio: Good catch, let's cut another release, doesn't hurt anything.

src/components/NWCBudgetEditor.tsx
```diff
@@ -0,0 +1,190 @@
...
+        initialValues: {
+            connection_name:
+                props.initialProfile?.name || props.initialName || "",
+            // If there's an initial profile, look at that, otherwise default to false
+            auto_approve: props.initialProfile
+                ? !props.initialProfile.require_approval
+                : false,
+            budget_amount:
+                props.initialProfile?.budget_amount?.toString() || "0",
+            interval: "Day"
```
benalleng, line 47: Last thing and then its perfect! This allows the interval to have the lastSelected value be displayed instead of only Day
``` suggestion
            interval: props.initialValues?.budget_period as BudgetForm["interval"] || "Day"
```

TonyGiorgio:
This should not be here for the amount editor on the budget input. 

Screenshot from 2023-09-18 21-43-16

TonyGiorgio (approved):
Really nice work, appreciate all of the iterations and fixes! Small nit on the amount screen, but otherwise this worked really well with my existing mainnet wallet. Profiles loaded that were already there, allowed me to edit them, add a budget, have those auto pay, all worked smoothly.

benalleng (approved):
Ditto, great work!

src/components/NWCBudgetEditor.tsx
futurepaul: nice thank you

futurepaul:
bumped to 0.4.19 (new commit because it's kind of orthogonal), removed the amount warning in the amount editor, and added that fix for the dropdown initial value

PR #503, opened 2023-08-31
Add option to select desired fiat currency

benalleng:
After https://github.com/MutinyWallet/mutiny-node/pull/736 we should be able to choose any currency we want

### TODO
- [ ] Tests?
- [ ] ~Choose better currency-loading design (If we even really want it)~ We can address in a different PR the current loading elipsis is fine for now

### DIFF
--- a/src/components/AmountCard.tsx
+++ b/src/components/AmountCard.tsx
@@ -3,7 +3,7 @@ import { createMemo, Match, ParentComponent, Show, Switch } from "solid-js";
 import { AmountEditable, Card, VStack } from "~/components";
 import { useI18n } from "~/i18n/context";
 import { useMegaStore } from "~/state/megaStore";
-import { satsToUsd } from "~/utils";
+import { satsToFormattedFiat } from "~/utils";
 
 const noop = () => {
     // do nothing
@@ -24,7 +24,6 @@ const KeyValue: ParentComponent<{ key: string; gray?: boolean }> = (props) => {
 export const InlineAmount: ParentComponent<{
     amount: string;
     sign?: string;
-    fiat?: boolean;
 }> = (props) => {
     const i18n = useI18n();
     const prettyPrint = createMemo(() => {
@@ -32,34 +31,34 @@ export const InlineAmount: ParentComponent<{
         if (isNaN(parsed)) {
             return props.amount;
         } else {
-            return parsed.toLocaleString();
+            return parsed.toLocaleString(navigator.languages[0]);
         }
     });
 
     return (
         <div class="inline-block text-lg">
             {props.sign ? `${props.sign} ` : ""}
-            {props.fiat ? "$" : ""}
-            {prettyPrint()}{" "}
-            <span class="text-sm">
-                {props.fiat ? i18n.t("common.usd") : i18n.t("common.sats")}
-            </span>
+            {prettyPrint()} <span class="text-sm">{i18n.t("common.sats")}</span>
         </div>
     );
 };
 
 function USDShower(props: { amountSats: string; fee?: string }) {
-    const i18n = useI18n();
     const [state, _] = useMegaStore();
-    const amountInUsd = () =>
-        satsToUsd(state.price, add(props.amountSats, props.fee), true);
+    const amountInFiat = () =>
+        (state.fiat.value === "BTC" ? "" : "~") +
+        satsToFormattedFiat(
+            state.price,
+            add(props.amountSats, props.fee),
+            state.fiat
+        );
 
     return (
         <Show when={!(props.amountSats === "0")}>
             <KeyValue gray key="">
                 <div class="self-end">
-                    ~{amountInUsd()}&nbsp;
-                    <span class="text-sm">{i18n.t("common.usd")}</span>
+                    {amountInFiat()}&nbsp;
+                    <span class="text-sm">{state.fiat.value}</span>
                 </div>
             </KeyValue>
         </Show>
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -1,5 +1,6 @@
 import { Dialog } from "@kobalte/core";
 import {
+    createEffect,
     createResource,
     createSignal,
     For,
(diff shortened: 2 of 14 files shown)


src/i18n/en/translations.ts
```diff
@@ -339,6 +338,18 @@ export default {
             forgot_password_link: "Forgot Password?",
             error_wrong_password: "Invalid Password"
         },
+        general: {
```
benthecarman, line 341: think this would be better at the top of this section

src/components/AmountEditable.tsx
```diff
@@ -187,13 +171,62 @@ export const AmountEditable: ParentComponent<{
...
     const [localFiat, setLocalFiat] = createSignal(
-        satsToUsd(
+        satsToFiat(
             state.price,
             parseInt(props.initialAmountSats || "0") || 0,
             false
         )
     );
 
+    const fixedAmount = (multiplier: number, label: boolean) => {
```
futurepaul, line 181: I'm confused by some of this multiplier logic... the fact that there are hardcoded matches in here makes me think this isn't a very generic function, but I'm scared that the logic will become even harder to follow if it's made generic somehow. maybe we should just discuss the best way to do it

src/components/layout/SelectField.tsx
futurepaul, line 1: we're long overdue for one of these!

src/utils/conversions.ts
```diff
@@ -12,20 +13,29 @@ export function satsToUsd(
...
             });
         } else {
             // Some float fighting shenaningans
-            const roundedUsd = Math.round(usd);
-            if (roundedUsd * 100 === Math.round(usd * 100)) {
-                return usd.toFixed(0);
+            const roundedFiat = Math.round(fiatPrice);
+            if (
+                roundedFiat * 100 === Math.round(fiatPrice * 100) ||
+                fiat?.match("KRW|JPY")
```
futurepaul, line 34: need to discuss the need for these hardcoded values doesn't seem like it will scale well to other currencies but also I don't want this logic any more complicated or my poor brain will fry!

src/state/megaStore.tsx
```diff
@@ -12,6 +12,7 @@ import {
 import { createStore } from "solid-js/store";
 import { useSearchParams } from "solid-start";
 
+import { showToast } from "~/components";
```
futurepaul, line 15: this store is kind of outside the "UI tree" so while it might work I'm not super comfy doing ui stuff like show toast from here. if there's a serious error it should set an error for the whole app, if it's not serious it should just fail gracefully

src/i18n/en/translations.ts
```diff
@@ -5,7 +5,7 @@ export default {
         home: "Home",
         sats: "SATS",
         sat: "SAT",
-        usd: "USD",
+        fiat: "{{fiat}}",
```
futurepaul, line 8: no need to translate if there's no translation!

futurepaul (changes requested):
this feels really close just want to see if we can simplify / explain some of the spicier logic in here for the sake of comprehension / future maintainability. probably worth doing a call before you try and rework it

futurepaul:
oh btw before I forget, these overflow on mobile:

Screenshot 2023-09-05 at 10 20 12 PM

src/components/AmountCard.tsx
```diff
@@ -42,7 +43,9 @@ export const InlineAmount: ParentComponent<{
             {props.fiat ? "$" : ""}
             {prettyPrint()}{" "}
             <span class="text-sm">
-                {props.fiat ? i18n.t("common.usd") : i18n.t("common.sats")}
+                {props.fiat
+                    ? i18n.t("common.fiat", { fiat: state.fiat })
```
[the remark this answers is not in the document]
benalleng: Yeah, trying to be too clever... I built this thinking they would change but there is an international standard for a reason.

src/state/megaStore.tsx
benalleng: Ok! At the moment it shows the toast and immediately switches to BTC as the pair

benalleng:
So for the overflow I changed it from being a simple .toFixed(8) to using maximumFractionalDigits: 8

image

src/components/AmountEditable.tsx
benalleng: Ok! I thought about doing it based on price but you might be right about even more added complexity there

benalleng:
The [support locale formats](https://github.com/MutinyWallet/mutiny-web/pull/503/commits/f11319d07e83485e1f23653ab22551500f5b01a7) commit allows for people with their browser language to view the numbers as they expect to see them in their country. Should not impact how any sats <-> fiat logic works

src/components/AmountEditable.tsx
benalleng: I changed the logic to use the price to determine a good fixedAmount to use, should allow for all currenies to have ~$1, ~$10, ~$100 option even if their price is a bit different compared to USD

src/components/AmountEditable.tsx
```diff
@@ -146,26 +180,55 @@ function SmallSubtleAmount(props: { text: string; fiat: boolean }) {
...
+        return parsed.toLocaleString(navigator.languages[0], {
+            minimumFractionDigits: 2
+        });
+    } else if (parsed === Math.trunc(parsed) && input.endsWith(".000")) {
+        return parsed.toLocaleString(navigator.languages[0], {
+            minimumFractionDigits: 3
+        });
+    } else if (
+        parsed !== Math.trunc(parsed) &&
+        input.match(/\.\d{2}0$/) &&
```
futurepaul, line 214: would like comments above non-obvious regexes

src/components/AmountEditable.tsx
```diff
@@ -146,26 +180,55 @@ function SmallSubtleAmount(props: { text: string; fiat: boolean }) {
...
+        input.includes(".", input.length - 4)
+    ) {
+        return parsed.toLocaleString(navigator.languages[0], {
+            minimumFractionDigits: 3
+        });
     } else if (
         parsed !== Math.trunc(parsed) &&
-        input.endsWith("0") &&
+        input.match(/\.\d{1}0$/) &&
```
futurepaul, line 222: comment plz

src/components/ChooseCurrency.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const handleFormSubmit = async (f: ChooseCurrencyForm) => {
+        setLoading(true);
+        try {
+            localStorage.setItem(
+                "fiat_currency",
+                JSON.stringify(findCurrencyByValue(f.fiatCurrency))
+            );
+
+            await timeout(1000);
+            window.location.href = "/";
```
futurepaul, line 126: is there a reason this requires a full reload rather than just a normal nav?

src/components/AmountEditable.tsx
```diff
@@ -378,23 +504,44 @@ export const AmountEditable: ParentComponent<{
...
+        setLocalFiat(satsToFiat(state.price, Number(sane) || 0, state.fiat));
     }
 
     function handleFiatInput(e: InputEvent) {
-        const { value } = e.target as HTMLInputElement;
-        const sane = fiatInputSanitizer(value);
+        const { value } = e.currentTarget as HTMLInputElement;
+        let sane;
+
+        if (state.fiat.value === "BTC") {
```
futurepaul, line 524: for big if else things like this I wouldn't mind some examples in the comments

src/components/AmountEditable.tsx
```diff
@@ -306,27 +400,57 @@ export const AmountEditable: ParentComponent<{
             character === "DEL" ||
             character === i18n.t("receive.amount_editable.del")
         ) {
-            if (localValue().length <= 1) {
+            if (
```
futurepaul, line 403: another case where there's enough logic here that some examples in the comments would be helpful

src/components/AmountEditable.tsx
```diff
@@ -187,13 +250,31 @@ export const AmountEditable: ParentComponent<{
...
+        satsToFiat(
             state.price,
             parseInt(props.initialAmountSats || "0") || 0,
-            false
+            state.fiat
         )
     );
 
+    const fixedAmount = (multiplier: number, label: boolean) => {
+        //This returns a stringified number based on the price range of the chosen currency as compared to BTC
```
futurepaul, line 261: I appreciate comments like this but they still require me to run a compiler in my head. even one example below this comment would be helpful for reading this

src/state/megaStore.tsx
```diff
@@ -74,6 +76,12 @@ export const Provider: ParentComponent = (props) => {
         deleting: false,
         scan_result: undefined as ParsedParams | undefined,
         price: 0,
+        fiat: {
```
futurepaul, line 79: I think you could do loading like this so you're not parsing every time in sync? just seems noisy code-wise:

```
        fiat: localStorage.getItem("currency") ? JSON.parse(localStorage.getItem("currency")!) as Currency : {
            value: "USD",
            label: "United States Dollar USD",
            hasSymbol: "$",
            maxFractionalDigits: 2
        },
```

then in sync you can use state.fiat knowing that it exists...

not a huge deal, I'm mostly just trying not to have the sync function overwhelmed by price-specific logic

futurepaul (approved):
this is really really good. it works super well even with all the complexity of the amount editor. my asks for comments are mostly selfishly for myself so I can understand it into the future

just a couple suggestions about saving / loading

src/state/megaStore.tsx
```diff
@@ -209,14 +217,41 @@ export const Provider: ParentComponent = (props) => {
...
+                            });
+                        } catch (e) {
+                            price = 1;
+                            setState({
+                                balance: newBalance,
+                                last_sync: Date.now(),
+                                price: price,
+                                fiat: {
+                                    value: "BTC",
+                                    label: "bitcoin BTC",
```
futurepaul, line 248: lowercase bitcoin here doesn't match the select option and it's the first thing you see when go to the chooser. could literally use `FIAT_OPTIONS[0]` if you wanted

src/components/ChooseCurrency.tsx
benalleng: I wasn't able to get actions.sync() to work to switch over to the new currency as a part of the async function

```
await actions.sync();
```
didn't really do anything so I just decided to reload from the homepage to force a sync

benalleng: I tried forcing a resync using
 
```
await actions.sync()
```

But nothing appeared to be happening, is there a better way to set a new state for the wallet?

futurepaul: normally what I do is create a setter in `actions`. here's how I save an npub for example:
```
saveNpub(npub: string) {
            localStorage.setItem("npub", npub);
            setState({ npub });
}
```

so in this case there could be `saveFiat` method that saves the new npub chosen here.

and then there could be a `fetchPrice` method that `sync` and `saveFiat` both call

benalleng: moved this TODO in #524

PR #510, opened 2023-09-04
Add channel reserve to lightning channels screen

### LINKED ISSUE #509: Add channel reserve to lightning channels page

benthecarman:
Closes #509

image

### DIFF
--- a/src/i18n/en/translations.ts
+++ b/src/i18n/en/translations.ts
@@ -245,11 +245,14 @@ export default {
             title: "Lightning Channels",
             outbound: "Outbound",
             inbound: "Inbound",
+            reserve: "Reserve",
             have_channels: "You have",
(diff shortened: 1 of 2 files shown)


benalleng (approved):
LGTM! Seeing that each bar is set to `min-w-fit` gives me a weird feeling that people may get confused about how much is actually in each section. But that is outside of the scope for this and not sure its all that important ultimately, just a visual UX thing.

benthecarman:
> Seeing that each bar is set to min-w-fit gives me a weird feeling that people may get confused about how much is actually in each section. But that is outside of the scope for this and not sure its all that important ultimately, just a visual UX thing.

Yeah think this is better way to do it. If the bars were the correct size no amount of text would be really visible

futurepaul (approved):
great improvement ty

PR #514, opened 2023-09-06
Fix downloading logs with a password set

### LINKED ISSUE #513: Can't download logs when password is set
```
2023-09-05 19:09:34.100 DEBUG [mutiny_wasm::indexed_db:200] Reading from local storage
Incorrect password entered.
```

Further, buttons like this should pop up with some error toast.

TonyGiorgio:
Fixes #513 

Requires https://github.com/MutinyWallet/mutiny-node/pull/751

### DIFF
--- a/src/components/Logs.tsx
+++ b/src/components/Logs.tsx
@@ -1,15 +1,30 @@
 import { MutinyWallet } from "@mutinywallet/mutiny-wasm";
+import { createSignal, Show } from "solid-js";
 
-import { Button, InnerCard, NiceP, VStack } from "~/components";
+import {
+    Button,
+    InfoBox,
+    InnerCard,
+    NiceP,
+    SimpleDialog,
+    TextField,
+    VStack
+} from "~/components";
 import { useI18n } from "~/i18n/context";
-import { downloadTextFile } from "~/utils";
+import { downloadTextFile, eify } from "~/utils";
 
 export function Logs() {
     const i18n = useI18n();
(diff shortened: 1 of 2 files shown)


src/components/Logs.tsx
```diff
@@ -1,25 +1,57 @@
...
+    const [error, setError] = createSignal<Error>();
+    const [exportDecrypt, setExportDecrypt] = createSignal(false);
+    const [password, setPassword] = createSignal("");
+
     async function handleSave() {
         try {
-            const logs = await MutinyWallet.get_logs();
-
+            setError(undefined);
+            const logs = await MutinyWallet.get_logs(password()); // Use password if required
```
futurepaul, line 27: I'm not totally sure what the diff between `get_logs("")` and `get_logs(undefined)` is (the other decrypt logic only runs when a password is needed, so it's always a string). but I'm pretty sure in case of a `None` you want `undefined`, not `""`

`const [password, setPassword] = createSignal();` will initialize `password()` to undefined. and you can clear it on line 47 `setPassword(undefined)`

TonyGiorgio: We handle both as the same because the empty string / undefined behavior is such a mess in javascript. So what you're suggesting is I just change one line? That's fine.

benthecarman:
CI failing

futurepaul (approved):
worked for me, can download logs with and without encryption

PR #516, opened 2023-09-06
Center X for nostr contact tags

benalleng:
I think this ultimately has to do with the fact that solid-select doesn't actually use an "x" but instead uses an "⨯" character so centering is rather difficult.

transitioning over to using a kobalte-core select like in the currency selector may be the move but for now, adjusting the negative margin centers the x

production:
image

PR:
image

### DIFF
--- a/src/styles/solid-select.css
+++ b/src/styles/solid-select.css
@@ -28,7 +28,7 @@
 
 .solid-select-multi-value-remove {
     /* TODO: there's gotta be a better way to vertically center this */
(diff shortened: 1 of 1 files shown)


TonyGiorgio (approved):
LGTM, will let paul judge whether or not this is fine to change

futurepaul (changes requested):
sadly this is different per browser and per platform (diff on chrome mac than chrome windows, diff on ff and diff on safari... annoying!)

PR #518, opened 2023-09-06
remove storybook and turbo

futurepaul:
talked with @benalleng and we agreed that storybook is not getting us the benefits we were looking for, and it's been a hassle (including breaking on node 20 due to some random babel dep)

@benalleng can you verify I got it all and put stuff back in the right place?


benalleng (approved):
LGTM! It brings a single tear to my eye knowing the time it took to set it up vs the time it took to revert it lol

PR #524, opened 2023-09-08
Currency change no longer requires reload

benalleng:
relies on https://github.com/MutinyWallet/mutiny-node/pull/758

works with the addition of mutiny-node#758 such that a new currency is fetched but switching back before the cache timeout keeps the currencies in cache so they can be swapped quickly without querying the coingecko api. The bitcoin currency option still never reaches out to the api.

Also makes the SmallSubtleAmount update when the price changes, this fixes a bug whereby the user could race the price api and set the sats amount before the fiat amount had a chance to be calculated. Production leaves the secondary amount at 0

[!](https://github.com/MutinyWallet/mutiny-web/assets/108441023/5723a9c4-06b3-47ac-bbd1-b52634e3430e)

### DIFF
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -283,6 +283,19 @@ export const AmountEditable: ParentComponent<{
         )
     );
 
+    const setSecondaryAmount = () =>
+        mode() === "fiat"
+            ? setLocalSats(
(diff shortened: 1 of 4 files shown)


futurepaul (approved):
nice, change looks clean and it works! double plus

PR #530, opened 2023-09-14
Add version to bottom of settings

benalleng:
@futurepaul Happy to make it a full clickable settingsLink if you think it should live separately

### DIFF
--- a/src/components/layout/ExternalLink.tsx
+++ b/src/components/layout/ExternalLink.tsx
@@ -3,9 +3,9 @@ import { ParentComponent } from "solid-js";
 export const ExternalLink: ParentComponent<{ href: string }> = (props) => {
     return (
         <a target="_blank" rel="noopener noreferrer" href={props.href}>
-            {props.children}{" "}
+            {props.children}
             <svg
-                class="inline-block"
+                class="inline-block pl-0.5"
                 width="16"
                 height="16"
                 fill="none"
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -6,6 +6,9 @@ import autoprefixer from "autoprefixer";
 import tailwindcss from "tailwindcss";
 
 import * as path from "path";
+import * as child from "child_process";
+
+const commitHash = child.execSync("git rev-parse --short HEAD").toString().trim();
 
 const pwaOptions: Partial<VitePWAOptions> = {
     base: "/",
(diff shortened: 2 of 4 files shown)


src/components/layout/ExternalLink.tsx
```diff
@@ -3,9 +3,9 @@ import { ParentComponent } from "solid-js";
 export const ExternalLink: ParentComponent<{ href: string }> = (props) => {
     return (
         <a target="_blank" rel="noopener noreferrer" href={props.href}>
-            {props.children}{" "}
+            {props.children}
             <svg
-                class="inline-block"
+                class="inline-block pl-0.5"
```
futurepaul, line 8: I don't think tailwind works like this. what were you trying to fix?

vite.config.ts
```diff
@@ -49,6 +52,10 @@ export default defineConfig({
         }
     },
     plugins: [wasm(), solid({ ssr: false }), VitePWA(pwaOptions)],
+    define: {
+        "import.meta.env.__COMMIT_HASH__": JSON.stringify(commitHash),
+        "import.meta.env.__RELEASE_VERSION__": JSON.stringify("0.4.16")
```
futurepaul, line 57: looks like there's a pretty nice way to get this direct from package.json. one less thing to update every version bump! https://stackoverflow.com/questions/67194082/how-can-i-display-the-current-app-version-from-package-json-to-the-user-using-vi

futurepaul (changes requested):
this is really slick. I think in the future we might want a whole "about" page where we can put more info but this is great for now. two comments on impl but concept ack

src/components/layout/ExternalLink.tsx
benalleng: spacing between the text and the externalLink children and the upArrow, right now the whitespace string adds an underline over nothing.

image

image

vite.config.ts
benalleng: well thats super easy

src/components/layout/ExternalLink.tsx
benalleng: No theres a 0.5 https://tailwindcss.com/docs/padding

futurepaul: wow TIL I am in fact crazy

futurepaul (approved):
thank you!

PR #534, opened 2023-09-15
Nostr activity and tag support

benalleng:
Testing autofill tags on contact send

[!](https://github.com/MutinyWallet/mutiny-web/assets/108441023/4a42db39-a21d-4965-b983-70e9dee32628)

### DIFF
--- a/src/components/ContactViewer.tsx
+++ b/src/components/ContactViewer.tsx
@@ -56,6 +56,7 @@ export function ContactViewer(props: {
                 showToast(result.error);
                 return;
             } else {
+                result.value.privateTag = props.contact.name;
                 if (
                     result.value?.address ||
                     result.value?.invoice ||
--- a/src/components/NostrActivity.tsx
+++ b/src/components/NostrActivity.tsx
@@ -6,10 +6,9 @@ import {
     Show,
     Switch
 } from "solid-js";
-import { Dynamic } from "solid-js/web";
 
 import rightArrow from "~/assets/icons/right-arrow.svg";
-import { AmountSats, TinyText, VStack } from "~/components";
+import { AmountSats, VStack } from "~/components";
 import { useI18n } from "~/i18n/context";
 import { useMegaStore } from "~/state/megaStore";
 import { fetchZaps, getHexpubFromNpub } from "~/utils";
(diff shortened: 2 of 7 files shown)


src/components/NostrActivity.tsx
```diff
@@ -151,10 +154,36 @@ export function NostrActivity() {
                         </div>
                         <Show when={zap.content}>
                             <hr class="my-2 border-m-grey-750" />
-                            <p
-                                class="truncate text-center text-sm font-light text-neutral-200"
-                                textContent={zap.content}
-                            />
+                            <TinyText>
```
futurepaul, line 157: this is nice but let's discuss / design it and do it in a separate PR

src/components/TagEditor.tsx
```diff
@@ -28,12 +29,24 @@ export function TagEditor(props: {
...
+                    ]);
+                }
+            }
         }
     });
 
     const selectProps = createMemo(() => {
         return createOptions(availableTags() || [], {
             key: "name",
+            disable: (value) => props.selectedValues.includes(value),
```
futurepaul, line 49: this is great thank you

src/components/TagEditor.tsx
```diff
@@ -70,10 +80,16 @@ export function TagEditor(props: {
             />
             <div class="flex flex-wrap gap-2">
```
futurepaul, line 81: what if we just drop this whole section? it's supposed to update with recommended tags as you use them, but right now it's just kind of confusing

src/components/layout/Misc.tsx
```diff
@@ -253,6 +253,7 @@ export const TinyText: ParentComponent = (props) => {
 export const TinyButton: ParentComponent<{
     onClick: () => void;
     tag?: MutinyTagItem;
+    hidden?: boolean;
```
futurepaul, line 256: I think the need for something like this is more of a symptom than a cure. let's just not have these buttons while we don't have a good way to get the correct list of them

src/components/ContactViewer.tsx
```diff
@@ -56,6 +56,7 @@ export function ContactViewer(props: {
                 showToast(result.error);
                 return;
             } else {
+                result.value.privateTag = props.contact.name;
```
futurepaul, line 59: this `privateTag` thing was a really slick solution

futurepaul (changes requested):
kind of like what you were saying on our call, I think it the functionality where you click "pay" on a contact and it autopopulates is awesome, the other stuff is more of a symptom of the tag editor being not-great right now. so I'd just do the payee tagging stuff (and keep the disable state for the dropdown list, that's a nice fix) and drop the rest for now

benalleng:
Going to correct the scope of this PR, It became sort of a test PR for me to demonstrate some concepts but we are going to move those things explored into #546 as they are more tag focused.

src/components/TagEditor.tsx
benalleng: Went ahead and completely removed it. maybe we should poll if we want it as is it remove it until the redesign

futurepaul (approved):
good stuff thank you

PR #539, opened 2023-09-19
revert: amount loading logic

benalleng:
Reverts a change from #503 so that there is not an amount loading indicator.

In combination with https://github.com/MutinyWallet/mutiny-node/pull/764 node version we should be able to never see a load time for instant prices


futurepaul:
not sure why cloudflare says still deploying, here's the url: https://dee5764b.mutiny-web.pages.dev/

futurepaul (approved):
it's dumb how much faster this makes the wallet feel lol

PR #543, opened 2023-09-20
fix safari app links

futurepaul:
there was an "open app" button that wasn't opening the uri callback because safari was being too aggressive about "pop-up blocking" (safari and other browsers don't want the browser to navigate if not by a user action). by lifting the logic out of capacitor's Browser.open safari is happy now.


benthecarman (approved):
CF didn't make a preview link but lgtm

TonyGiorgio:
I'm pretty sure I had tested this when I swapped it over but maybe not.

TonyGiorgio (approved):

PR #544, opened 2023-09-20
Add PWA shortcuts for send, receive, and activity

benthecarman:
~~Havent tested this yet (not sure if you can with iOS) but wanted to play around with it~~

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Expose_common_actions_as_shortcuts


benthecarman:
Got it working for android, needed to use dev version of chrome, can't figure out why the icons aren't showing up. Might be my manifest isn't updating..

https://web.dev/manifest-updates/#cr-desktop-test

image

benthecarman:
Also works on chrome desktop
image

benalleng:
tACK I can see the various shortcuts but I also cannot see the icons. What if you change their fill/background? I noticed on other apps the background is always white. maybe the icons are there, they are just not visible

benthecarman:
> Is there a way to add these shortcuts to the android build? Or does that become more of a rabbit-hole of widgets and stuff

looks like we can use this with capacitor 

https://github.com/NePheus/capacitor-android-shortcuts

benalleng:
That worked!

Screenshot_20230921-182516

benthecarman:
Sweet how should we color these then

futurepaul:
activity
receive
send

benalleng:
Tack

TonyGiorgio:
Still not working for me with vanadium on Android but it's not hurting anything at least.

futurepaul (approved):
lol so many files. lgtm def checks a lot more boxes on that pwabuilder test

PR #558, opened 2023-09-24
Only optimistically get price when balance non-zero

benthecarman:
When it is a fresh wallet they won't have a cached price so it'll be slow to get the price, so we can skip getting the price and just let the interval do it.


futurepaul (approved):
nice, thank you

PR #559, opened 2023-09-25
add vss config option

futurepaul:
used this opportunity to improve how we handle setup errors outside the main route. before it would hang if error when trying to load `/settings/servers`. still not a PERFECT story for if someone puts in a bad url and saves, but they should still be able to nav to settings and fix. I made it so settings doesn't even rely on megastore which should make it extra resilient. could potentially add a link on an error page but would only want to do that if we were confident that's the right move


.github/workflows/android-prod.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           VITE_RGS: https://scorer.mutinywallet.com/v1/rgs/snapshot/
           VITE_AUTH: https://auth.mutinywallet.com
           VITE_SUBSCRIPTIONS: https://subscriptions.mutinywallet.com
-          VITE_STORAGE: https://storage.mutinywallet.com
+          VITE_STORAGE: https://storage.mutinywallet.com/v2
```
TonyGiorgio, line 67: There's many more places this needs to also be updated.

TonyGiorgio (changes requested):

.github/workflows/android-prod.yml
benthecarman: Ah, i missed the staging urls

TonyGiorgio: Go ahead and apply to cloudflare now so we don't forget. it won't apply until a release, though it may break staging stuff in other PR's until then, but we need to test it in this branch.

.env.example
```diff
@@ -14,7 +14,7 @@ VITE_RGS="https://rgs.mutinynet.com/snapshot/"
 VITE_SELFHOSTED="true"
 VITE_AUTH="https://auth-staging.mutinywallet.com"
 VITE_SUBSCRIPTIONS="https://subscriptions-staging.mutinywallet.com"
-VITE_STORAGE="https://storage-staging.mutinywallet.com"
+VITE_STORAGE="https://storage-staging.mutinywallet.com/v2"
```
TonyGiorgio, line 17: just double checking, is this the only ENV change that needs to be added to `.env.example` ?

benthecarman: yeah, nothing else changed

TonyGiorgio (approved):

PR #563, opened 2023-09-27
Update example storage url

benthecarman:
With https://github.com/MutinyWallet/mutiny-node/pull/769 merged we now need to use the `/v2` routes to stay compatible.


benthecarman:
in

Issue #564, opened 2023-09-28
"Channel reserve too high" message


Synesso:
This seems to happen when I try to sweep all my funds out. Payments that leave dust are sometimes successful (putting aside frequent "timeouts"), but it is not clear how much I can sweep or why it fails.

benthecarman:
In settings there is a `Lightning Channels` page that will show how much you have available and what the reserve amount is

benthecarman:
@futurepaul maybe on the error screen for this we should display those balances?

Synesso:
For me that page shows

outbound: 20,110
reserve: 546
inbound: 121,305

But I receive the Channel reserve too high message when attempting to send
amounts between 19k and 20k.

benthecarman:
Looks like our outbound number is incorrect, that isn't removing the reserve amount

benthecarman:
Generally trying to send the exact amount of outbound isn't possible, fees will make it hard to predict exactly how much you can get out. If you're looking to sweep your balance normally your best bet is to close the channel and send on-chain.

However I would put in a word of caution, right now we have an issue when you close a channel the fee rate is very low so it may take awhile for you to get your sats. Our next release (sometime this week) will fix this.

TonyGiorgio:
> @futurepaul maybe on the error screen for this we should display those balances?

Yes, very much needed. We need to be able to distinguish failed payments from routing vs failed payments simply because the amount was too high. Perhaps also a warning when trying to send close to 1% that it might fail due to reserrve limits.

TonyGiorgio:
Oh nevermind, I didn't realize we already had that specific error page. Yeah I guess this would fix this issue a little bit, just tested it and it's accurate, though still a bit of a problem getting too close to fees: https://github.com/MutinyWallet/mutiny-web/pull/565

PR #565, opened 2023-09-28
Correct outbound calculation

benthecarman:
Before we were showing our ln balance as outbound, however for our LN balance we include things like the reserve so it doesn't look like you randomly lose money. This fixes so we actually correctly calculate outbound capacity

Related to #564


TonyGiorgio:
So this is just applied to the settings page? Currently it's double counting reserve, once in outbound and once in the new reserve balance?

benthecarman:
Yeah just on the LN channels page

TonyGiorgio (approved):
tACK

PR #566, opened 2023-09-28
Trim pasted seed words

benthecarman:
Was trying to paste my seed words but getting errors with a trailing space


TonyGiorgio (approved):
Ran into that once too

PR #571, opened 2023-09-29
hide mutiny+ if self-hosting

futurepaul:
is there other stuff that we should modify / hide for self-hosted?


benthecarman (commented):
Should add a tag line at the top (similar to `Signet`) for self hosting

benthecarman:
CI failing

benthecarman (approved):

PR #572, opened 2023-10-01
Payjoin sends

benalleng:
Relies on https://github.com/MutinyWallet/mutiny-node/pull/647 and https://github.com/MutinyWallet/bitcoin-waila/pull/8

Test with @DanGould's payjoin receiver 
`bitcoin:TB1QSAQ38TWXQEJGR3JJ3AHGHWLSMV8JJCFDWERGM4?pj=https://mutiny.payjoin.org:4433`

- [x] waila part
- [x] node part

### DIFF
--- a/src/logic/waila.ts
+++ b/src/logic/waila.ts
@@ -7,6 +7,7 @@ await initWaila();
 
 export type ParsedParams = {
     address?: string;
+    pj_uri?: string;
     invoice?: string;
     amount_sats?: bigint;
     network?: string;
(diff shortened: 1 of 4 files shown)


src/logic/waila.ts
```diff
@@ -49,6 +50,7 @@ export function toParsedParams(
         ok: true,
         value: {
             address: params.address,
+            pj_uri: params.pj_uri,
```
benthecarman, line 53: instead of doing this how I think we are gonna handle it is have it called `payjoin_enabled` and we set that as `params.payjoin_endpoint !== undefined`

then when we call `send_payjoin` we use the original input string. that way we can just let pdk handle it

benthecarman: I just published waila v0.2.3 that has this function

benalleng:
Closed in favor of #608

PR #573, opened 2023-10-01
Handle Copy for http context

benthecarman:
recommended from here: https://github.com/getumbrel/umbrel-apps/pull/785#issuecomment-1741693788


ParthJadhav (approved):

benthecarman:
Many other features only work with TLS so no real reason to support this. Closing

PR #575, opened 2023-10-01
Currency selector no longer resets to USD on no-selection

benalleng:
When pressing the select currency button without selecting an option, it would always reset to USD even if the displayed select option value was another currency

I was able to use state.fiat as the logic for setting the fiat will be done by the time the user gets to this page, no concern of undefined behavior


futurepaul (approved):
nice catch ty, didn't notice this

PR #584, opened 2023-10-04
Update needed for ldk-117

benthecarman:
with ldk 117 I added the ability to add a message to keysends, this just fixes the api call and puts a todo for us to eventually add

needs https://github.com/MutinyWallet/mutiny-node/pull/772/


TonyGiorgio (approved):
LGF

PR #587, opened 2023-10-06
DX Improvement: Adds Husky for automatic linting/formatting/typechecking on commits

thebrandonlucas:
Currently you have to `pnpm run pre-commit` manually before commits, which can be easy to forget and is one extra step. This PR adds [Husky](https://typicode.github.io/husky/getting-started.html), which is a tool that automatically runs before every `git commit` (or whatever other git commands you would like it to run before),

In this case, I automated the `pnpm run pre-commit` step, and removed the comment from the README requiring the user to run that before every commit.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "com.mutinywallet.mutinywallet"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 31
-        versionName "0.4.30"
+        versionCode 32
+        versionName "0.4.31"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
              // Files and dirs to omit from the packaged assets dir, modified to accommodate modern web apps.
--- a/ios/App/App.xcodeproj/project.pbxproj
+++ b/ios/App/App.xcodeproj/project.pbxproj
@@ -357,7 +357,7 @@
 				INFOPLIST_KEY_LSApplicationCategoryType = "public.app-category.finance";
 				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				MARKETING_VERSION = 1.4.30;
+				MARKETING_VERSION = 1.4.31;
 				OTHER_SWIFT_FLAGS = "$(inherited) \"-D\" \"COCOAPODS\" \"-DDEBUG\"";
 				PRODUCT_BUNDLE_IDENTIFIER = com.mutinywallet.mutiny;
 				PRODUCT_NAME = "$(TARGET_NAME)";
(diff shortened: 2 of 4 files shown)


futurepaul (changes requested):
Hey brandon, happy to see you here!

I tried this out and it made the commit and THEN the changes were applied afterward (so I got a bad commit logged). I think ideally this should fail to commit if format or typecheck fails, and then we can have a diff command for fixing formatting. (Also would be good to say in readme how to skip the check if you don't want it for some commit, hate to be locked in can't-commit prison)

thebrandonlucas:
Hey Paul, sorry for the mistake, the doesn't-apply-formats-to-commit issue is now fixed (I was forgetting to re-stage the format changes in husky's `pre-commit` file).

Regarding failing, it's running the `pre-commit` script in `package.json`, so it fails on TS errors (not on warnings -- this can be configured but I thought it's probably too strict) and `eslint` errors, but not on prettier errors, which it just auto-formats, which makes the most sense to me from an optimal DX perspective (it's probably very rare that a dev would want to see _how_ prettier is formatting and manually run the format script instead of just having it do it for them), but that's just my opinion; happy to change it to `pnpm run check-format` if you'd prefer.

Also updated the README to reflect the new changes and tell the user how to escape commit prison.

Let me know if there's anything else I need to do before merging!

.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+pnpm run pre-commit
+git add --all
```
futurepaul, line 5: if someone is trying to just commit one file or one small change, won't this auto-commit all changes? I think yeah probably just want to do `pnpm run check-format`. I'd much rather error out of a commit than have all my changes auto committed without my intent.

futurepaul (changes requested):

.husky/pre-commit
thebrandonlucas: Ah, thank you for this catch, I looked into this a bit more and found [lint-staged](https://www.npmjs.com/package/lint-staged), which only applies the the linting/type-checking/formatting to **staged** files, runs the jobs in parallel, and only displays output on error (I'm sure it can be configured if we want warnings as well). Paired with Husky this is the optimal DX afaict. It's clean, configurable, and catches mistakes up front very nicely.

Here's an example of an error output with the following code:
Screenshot 2023-10-20 at 12 19 27 PM

Screenshot 2023-10-20 at 12 21 45 PM

A non-error output:
Screenshot 2023-10-20 at 1 00 11 PM

README.md
```diff
@@ -6,7 +6,7 @@
 
 ```
 pnpm install
-pnpm run dev
+pnpm dev
```
thebrandonlucas, line 9: removed unnecessary `run` command

.husky/pre-commit
futurepaul: yeah I think this is a good improvement. now it just needs to work! I'm getting a bunch of wild output, do you know why this would be? (this commit should be good, I only added a comment)

Screenshot 2023-10-20 at 4 30 40 PM

thebrandonlucas: I wasn't getting this output for some reason until I reran `pnpm install`. Weird. Anyway the solution ended up having to do with typescript not being able to find the `tsconfig.json` when invoked from husky, but when we run it as bash it can: `bash -c 'pnpm run check-types'`. Should be fixed now.

More here: https://github.com/lint-staged/lint-staged/issues/825#issuecomment-674575655

futurepaul (approved):
okay sorry for the delay but this seems to be behaving great for me now!

please rebase and turn this into a single commit and we're good to go

PR #595, opened 2023-10-13
Display all send fieldDestination errors

### LINKED ISSUE #594: No feedback when failing an attempt to send to a lightning address
Using the web wallet on Chrome on Android, I attempt to send to a lightning address by clicking send then put an address in the form name@domain.com

When I click Continue, the button briefly turns blue, but nothing happens. No explanation what error is occurring. The lightning address stays in the field, all I can do is go back.

benalleng:
Closes #594 
Should catch more destination parsing errors relating to things beyond the waila

(also moves some component child-parent organization so that buttons and errors aren't visible before they should be

Something to think about, should consider making a more "user" readable error vs human readable error here?

### TODO
- more testing to make sure I've caught them all

### DIFF
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -364,6 +364,9 @@ export default function Send() {
                             setLnurlp(source.lnurl);
                             setSource("lightning");
                         }
+                    })
+                    .catch((e) => {
+                        showToast(eify(e));
                     });
             } else {
                 setAmountSats(source.amount_sats || 0n);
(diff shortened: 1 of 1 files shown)


futurepaul (changes requested):
I'll need to dig more into this to know the correct move here but this logic isn't right. I can end up in an amount-setting screen with no continue button:

Screenshot 2023-10-17 at 2 06 06 PM

futurepaul (approved):
thank you for your service

PR #598, opened 2023-10-16
Vibrate on successful send, receive, channel open

### LINKED ISSUE #588: Vibrate on success send and receive
Not sure if we can do this for the web but might be able to for the native apps

benthecarman:
Closes #588

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -14,6 +14,7 @@ dependencies {
     implementation project(':capacitor-clipboard')
     implementation project(':capacitor-filesystem')
     implementation project(':capacitor-share')
+    implementation project(':capacitor-haptics')
     implementation project(':capacitor-toast')
     implementation project(':mutinywallet-barcode-scanner')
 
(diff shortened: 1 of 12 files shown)


benalleng:
Doesn't seem to work on Android Chrome or Firefox. Either installed as PWA or just on the Web. I noticed that it did work on Android if the version duration was longer when I tried it on whatpwacando

benthecarman:
upped to 1 second and still not working for me :/

benthecarman:
This is working for me on one of my androids but not another. Not sure if that is just browsers being dumb.

benalleng:
tAck on the apk build works fine there, just another reason to download it i guess.

futurepaul (approved):
going to leave this to you guys for testing because I don't have a handy android setup right now, but the code looks good / straightforward. I'll chime in more on this when I start doing the ios stuff, guessing this should work for the ios app (obvi not working in ios pwa)

benalleng (approved):
Figured out my issue, code works well

TonyGiorgio:
Have you ran `just native` yet? There's typically android changes whenever capacitor packages are added but I don't see any android change.

benthecarman:
> Have you ran `just native` yet? There's typically android changes whenever capacitor packages are added but I don't see any android change.

Added that, however it was working without it 🤷‍♂️ . We did already have the permissions for vibration I guess

PR #602, opened 2023-10-16
Error catch receive amount too low

### LINKED ISSUE #600: Better error message when first receive is under 10k
for a wallet without any inbound if you try to request less than 10k you get a too-vague error and just an on-chain address. Should be more descriptive and maybe not even a toast, but an actual InfoBox. Something like `Amount is too small for your initial Lightning receive.`
Screenshot 2023-10-16 at 12 54 50 PM

benalleng:
Closes #600 

~~Should I make this error catch as granular as it is or should I just return all possible errors from this particular catch like so and then log out the specifics? Don't know how else the logic above this catch could fail...~~ I added some new error messages for other possible failures

### DIFF
--- a/src/i18n/en/translations.ts
+++ b/src/i18n/en/translations.ts
@@ -69,6 +69,12 @@ export default {
         spendable: "Spendable",
         channel_size: "Channel size",
         channel_reserve: "- Channel reserve",
+        error_under_min_lightning:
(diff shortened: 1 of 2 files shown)


benthecarman (approved):
good change!

futurepaul (approved):
Nice, thank you!

PR #609, opened 2023-10-20
nwc callback url and url budget params

futurepaul:
given a url like this:

http://app.mutinywallet.com/settings/connections?return_to=RETURN_URL&name=CONNECTION_NAME&budget_renewal=INTERVAL&max_amount=BUDGET_AMOUNT

after adding the budget item it will go to the RETURN_URL. also if the user picks auto-approve it will have the budget and interval pre-filled

params are intentionally similar to https://github.com/getAlby/nostr-wallet-connect#client-created-secret


src/components/NWCBudgetEditor.tsx
```diff
@@ -31,29 +31,49 @@ export type BudgetForm = {
...
+    const connection_name = props.initialProfile?.name
+        ? props.initialProfile?.name
+        : props.initialName
+        ? props.initialName
+        : "";
+
+    // If there's an initial profile, look at that, otherwise default to false
+    const auto_approve = props.initialProfile?.require_approval
+        ? !props.initialProfile?.require_approval
+        : false;
```
benthecarman, line 51: Think this can just be `props.initialProfile?.require_approval || false`

src/components/NWCBudgetEditor.tsx
```diff
@@ -31,29 +31,49 @@ export type BudgetForm = {
...
 }) {
     const i18n = useI18n();
 
+    console.log(props);
+
+    const connection_name = props.initialProfile?.name
+        ? props.initialProfile?.name
+        : props.initialName
+        ? props.initialName
+        : "";
```
benthecarman, line 46: same here, think this can be simplified as 

`props.initialProfile?.name || props.initialName || "`

src/routes/settings/Connections.tsx
```diff
@@ -50,6 +50,24 @@ function mapIntervalToBudgetPeriod(
...
+        case "day":
+            return "Day";
+        case "week":
+            return "Week";
+        case "month":
+            return "Month";
+        case "year":
+            return "Year";
+        default:
+            return "Day";
```
benthecarman, line 67: is having a default safe here? think it might be better to just return undefined

src/components/NWCBudgetEditor.tsx
futurepaul: no there's a weird double negative here

futurepaul:
okay this logic shit is in a better state now

benthecarman (approved):

PR #611, opened 2023-10-20
Better deny all nwc

benthecarman:
Needs https://github.com/MutinyWallet/mutiny-node/pull/801

makes it so we do deny all in a batch


futurepaul (approved):
works for me, and I tested the basics just in case because of flaky test

PR #616, opened 2023-10-23
include nwc string in url callback

futurepaul:
also made it so auto approve is checked by default if there's an amount or interval in the incoming search params


futurepaul:
if you want to test this you can use https://www.zapplepay.com/autozap/

just replace `app.mutinywallet.com` urls with `fix-nwc-callback.mutiny-web.pages.dev` and it should all work good

benthecarman (approved):

benthecarman (approved):

PR #619, opened 2023-10-25
Use Capacitor for all vibrations

benthecarman:
when calling `window.navigator.vibrate` on browsers that didn't support it it would throw an error, so we just let Capacitor handle it for us


futurepaul (changes requested):
actually it's worse than we thought. on safari it doesn't show the receive screen due to the unhandled error (including with this version, apparently Capacitor isn't checking for the feature either)

Screenshot 2023-10-24 at 7 43 41 PM

futurepaul:
okay now it catches the error and just prints a warning

benthecarman (commented):
ack

PR #630, opened 2023-10-26
bump the z-index of amount editor

futurepaul:
have a problem using the amount editor on the wallet connection screen on safari. my hunch is it has something to do with nested dialogs (the amount editor is a modal "dialog" launched by another modal "dialog") I can't repro unfortunately on localhost but want to see if this works when deployed...


benthecarman (approved):
fixes for me

PR #635, opened 2023-10-27
Periodically check if we have inflight payments

benthecarman:
This will only work on chrome android but it should help with stuck htlc force closures. Every hour it'll check our storage and if we have an inflight payment, it'll give a push notification

https://developer.mozilla.org/en-US/docs/Web/API/Web_Periodic_Background_Synchronization_API


benthecarman:
hmm do we need to update nodejs version or something, not sure why it says periodic sync doesn't exist

futurepaul:
really smart solution for android. we could also probably do something similar to breeze on ios where the notification server just lets them know they need to check on their wallet at a regular interval if they haven't loaded it in over a week

some stuff to figure out that I would like to be helpful with:
1. we need to make sure this will get updated when we want to update it (browsers love to cache service workers)
2. need to make sure this doesn't get cached and make dev hell (basically the same as above, but for dev mode)
3. need to work this into vite build steps to get typescript right (which will also be important for when we want to use wasm in a service worker). I had that working in my test repo, but when I tried to do the same thing in https://github.com/MutinyWallet/mutiny-web/tree/service-worker-tmp-paul/src/service-worker I was getting typescript errors ... but for starters putting a .ts file in `public` doesn't make sense because it's not a static file it needs to be transformed to .js before it ships to browser

benthecarman:
We can also do this with the android app, that might be an easier first approach

https://capacitorjs.com/docs/apis/background-runner

PR #638, opened 2023-10-28
remove nostr dev kit dep

futurepaul:
needs https://github.com/MutinyWallet/mutiny-node/pull/812


src/utils/nostr.ts
```diff
@@ -0,0 +1,61 @@
...
+    CategorizedHighlightList = 39802,
+    DVMJobFeedback = 65000,
+    DVMJobResult = 65001,
+    DVMJobRequestTranscription = 65002
+}
+
+export async function hexpubFromNpub(
+    npub: string
+): Promise<string | undefined> {
+    if (!npub.startsWith("npub")) {
```
benthecarman, line 50: Npubs are case insensitive, should handle that here

futurepaul: fixed

benthecarman (approved):

PR #647, opened 2023-10-31
Prod v0.4.27

TonyGiorgio:


futurepaul (approved):
I missed these changes to android-staging workflow but i trust you on that

PR #648, opened 2023-10-31
Fix activity items with no fees

TonyGiorgio:


benthecarman (approved):
ACK for now

imo we should probably not show the fee if there isn't one but lets fix the bug

benthecarman:
needs `pnpm format`

benthecarman (approved):

PR #651, opened 2023-11-01
Add index html

futurepaul:
saw on solid discord apparently you can just throw in an index.html to the root and that will get used for static spa builds (what we do)

want to see if this improves our meta data on https://renderform.io/tools/twitter-card-preview-tool/#


futurepaul:
before:
Screenshot 2023-11-01 at 2 09 30 PM

after:
Screenshot 2023-11-01 at 2 09 13 PM

benthecarman:
You have a commit from another pr

benthecarman (approved):

futurepaul:
@TonyGiorgio you added `insertHead.js` so want to make sure this is accomplishing what that was... the android app seems to work fine for me in the emulator with this change

TonyGiorgio:
Do a full end to end test. Sadly I didn't comment on why it was needed. Might need to search back. It isn't initially needed but I'm pretty sure some things were broken. May I ask why you are removing it?

futurepaul:
it sounds related to this: https://stackoverflow.com/questions/62125279/what-does-unable-to-inject-capacitor-mean-and-how-can-i-fix-it

all this script does is add a `<head></head>` tag to the html during the build step so capacitor has a place to inject itself. but now that we have an html file that already has a `<head>` tag in it I don't think that's necessary?

TonyGiorgio:
Perfect, sounds good to me then.

TonyGiorgio (approved):

PR #656, opened 2023-11-02
Make MegaStoreProvider higher in root components

benthecarman:
No totally sure what I am doing here but I think this could help.

When `MegaStoreProvider` is mounted it creates an instance of mutiny-node, so if any parent components are changed it would too. So I put the `I18nProvider` inside of it incase that is changing as well as a second `ErrorBoundary` that will handle the `Router`, so if any uncaught errors happen there, we don't create a new instance of mutiny-node.

ChatGPT said it might help

### DIFF
--- a/src/components/ErrorDisplay.tsx
+++ b/src/components/ErrorDisplay.tsx
@@ -39,7 +39,9 @@ export function ErrorDisplay(props: { error: Error }) {
                 </NiceP>
                 <NiceP>
                     {i18n.t("error.general.getting_desperate")}{" "}
-                    <A href="/emergencykit">{i18n.t("error.emergency_link")}</A>
+                    <A href="/settings/emergencykit">
+                        {i18n.t("error.emergency_link")}
+                    </A>
                 </NiceP>
                 <div class="h-full" />
                 <Button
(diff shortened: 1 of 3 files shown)


benalleng:
The additional `ErrorBoundary` seems to be the change pulling the weight, when trying to induce a failure I could not cause a i18n error whether the `I18nProvider` was in the `MegaStoreProvider` or not. otherwise tACK on the change, errors now direct to a correct error page instead of the app failing

benalleng:
I did notice there is something wrong with `ErrorDisplay.tsx` the emergency kit link is incorrect, needs to be "/settings/emergencykit".
figure its small enough to push along with this
```
<NiceP>
    {i18n.t("error.general.getting_desperate")}{" "}
    <A href="/settings/emergencykit">{i18n.t("error.emergency_link")}</A>
</NiceP>
```

futurepaul (approved):
nice catch on the broken link @benalleng, fixed in a second commit 

I thought there was a specific reason I nested things how they're nested, but I don't see a direct dependence between i18n and megastore so I think it's fine either way. but either way the error boundary directly outside the router seems to be a huge win, thank you @benthecarman (would've caught that recent bug with the missing fee, for instance)

TonyGiorgio:
> I did notice there is something wrong with `ErrorDisplay.tsx` the emergency kit link is incorrect, needs to be "/settings/emergencykit". figure its small enough to push along with this
> 
> ```
> <NiceP>
>     {i18n.t("error.general.getting_desperate")}{" "}
>     <A href="/settings/emergencykit">{i18n.t("error.emergency_link")}</A>
> </NiceP>
> ```

I think this fixes an issue someone has seen with an emergency page last week

futurepaul:
Screenshot 2023-11-02 at 12 57 13 PM

PR #667, opened 2023-11-05
Hide lightning swap button if on-chain balance is only force closes

benthecarman:
A lot of users get confused after a force close they click the swap button but then have 0 balance there. We should only show the button if they have an actual balance, not pending force closes


futurepaul (approved):

futurepaul (approved):
lgtm but needs formatting apparently

PR #673, opened 2023-11-08
handle no js and lockdown mode case

### LINKED ISSUE #290: Do some Static Rendering so we get some minimal page for JS-less browsers

futurepaul:
this should handle the case where js is disabled (noscript tag), or if lockdown is enabled (js works, but wasm doesn't, so `no-wasm` tag)

Closes #290


benthecarman (approved):
Works for me

PR #676, opened 2023-11-08
Full coingecko currencies

### LINKED ISSUE #665: Request support for more currencies
I want the Norwegian krone as an option

benalleng:
Closes #665 

Adds all available currencies currently supported by our coingecko api, If we want to support more, we'll need to go the multi-api route. It shouldn't be too hard as we can append the correct api to search in each currency item in FIAT_OPTIONS


src/components/ChooseCurrency.tsx
futurepaul, line 1: the number of options causes the select menu to break on mobile. I think we should just use native select dialogs and get rid of the SelectField component (I already did this for NWC and meant to suggest it for this one). I think this is the code you need:

```tsx
<Field name="fiatCurrency">
    {(field, props) => (
        <select
            {...props}
            value={field.value}
            class="w-full rounded-lg bg-m-grey-750 py-2 pl-4 pr-12 text-base font-normal text-white"
        >
            <For each={FIAT_OPTIONS}>
                {({ value, label }) => (
                    <option value={value}>{label}</option>
                )}
            </For>
        </select>
    )}
</Field>
```

src/utils/currencies.ts
```diff
@@ -0,0 +1,293 @@
...
+        hasSymbol: "₫",
+        maxFractionalDigits: 0
+    },
+    {
+        label: "South African Rand ZAR",
+        value: "ZAR",
+        hasSymbol: "R",
+        maxFractionalDigits: 2
+    }
+].sort((a, b) => (a.value > b.value ? 1 : b.value > a.value ? -1 : 0));
```
futurepaul, line 290: this list looks already alphabetical, I think we can drop the sort

src/utils/currencies.ts
```diff
@@ -0,0 +1,293 @@
...
+
+export const USD_INDEX = FIAT_OPTIONS.findIndex((fo) => fo.value === "USD");
```
futurepaul, line 292: can hardcode these now if we get rid of sort. just make sure there's a good comment warning.

(I don't know exactly when these get re-run... would rather "compile time" compute them, rather than wasting cycles finding values in a list over and over)

futurepaul (changes requested):

src/utils/currencies.ts
benalleng: What do you think makes more sense, value sorted or label sorted?

src/state/megaStore.tsx
```diff
@@ -272,7 +279,12 @@ export const Provider: ParentComponent = (props) => {
                             balance: newBalance,
                             last_sync: Date.now(),
                             price: 1,
-                            fiat: FIAT_OPTIONS[BTC_INDEX]
+                            fiat: {
+                                label: "Bitcoin BTC",
+                                value: "BTC",
+                                hasSymbol: "₿",
+                                maxFractionalDigits: 8
+                            }
```
benthecarman, line 287: How about we make this a `const` in currencies.ts and you cant use it here and put it in the array. That way we get the best of both worlds, safety and no duplication

src/state/megaStore.tsx
```diff
@@ -97,7 +99,12 @@ export const Provider: ParentComponent = (props) => {
         price: 0,
         fiat: localStorage.getItem("fiat_currency")
             ? (JSON.parse(localStorage.getItem("fiat_currency")!) as Currency)
-            : FIAT_OPTIONS[USD_INDEX],
+            : {
+                  label: "United States Dollar USD",
+                  value: "USD",
+                  hasSymbol: "$",
+                  maxFractionalDigits: 2
+              },
```
benthecarman, line 107: same for USD

benthecarman (changes requested):

src/state/megaStore.tsx
benalleng: Went with a concat option with USD+BTC thus they are separate from the rest of the `FIAT_OPTIONS` and I put them at the front so that they are the most easily accessible options fro the user

src/utils/currencies.ts
benalleng: went with value sorted for now

src/components/ChooseCurrency.tsx
```diff
@@ -1,119 +1,36 @@
...
     const i18n = useI18n();
     const [error, setError] = createSignal<Error>();
-    const [state, actions] = useMegaStore();
+    const [_state, actions] = useMegaStore();
     const [loading, setLoading] = createSignal(false);
     const navigate = useNavigate();
 
     function findCurrencyByValue(value: string) {
-        return FIAT_OPTIONS.find((currency) => currency.value === value);
+        return (
```
benalleng, line 30: reverted this previous change from #575 realized that behavior was not ideal

benalleng: Tried to get the button disabled when `!chooseCurrencyForm.touched` but it wasn't working properly

src/state/megaStore.tsx
```diff
@@ -97,7 +97,7 @@ export const Provider: ParentComponent = (props) => {
         price: 0,
         fiat: localStorage.getItem("fiat_currency")
             ? (JSON.parse(localStorage.getItem("fiat_currency")!) as Currency)
-            : FIAT_OPTIONS[USD_INDEX],
+            : BTC_USD_OPTIONS[0],
```
futurepaul, line 100: I really don't like this hardcoded indexing into an array it's just hard to read at the callsite and lame to maintain.

how I interpreted @benthecarman's suggestion:

in currencies.ts make a `export const USD_CURRENCY` item, `export const BTC_CURRENCY` item, then the currencies array is `[BTC_CURRENCY, USD_CURRENCY, ...rest]`. now you can import BTC or USD items as values instead of indexing them. and there's no need to concat or sort the array

futurepaul (changes requested):

src/components/ChooseCurrency.tsx
futurepaul: no worries disabled buttons states are overrated

futurepaul (approved):
looks good to me, thank you for tackling!

src/components/ChooseCurrency.tsx
```diff
@@ -1,119 +1,37 @@
...
-        value: "KRW",
-        hasSymbol: "₩",
-        maxFractionalDigits: 0
-    },
-    { label: "Kuwaiti Dinar KWD", value: "KWD", maxFractionalDigits: 3 }
-].sort((a, b) => (a.value > b.value ? 1 : b.value > a.value ? -1 : 0));
-
-export const USD_INDEX = FIAT_OPTIONS.findIndex((fo) => fo.value === "USD");
-export const BTC_INDEX = FIAT_OPTIONS.findIndex((fo) => fo.value === "BTC");
+const COMBINED_OPTIONS: Currency[] = [USD_OPTION, BTC_OPTION, ...FIAT_OPTIONS];
```
benthecarman, line 21: I was thinking of putting `USD_OPTION, BTC_OPTION` inside of `FIAT_OPTIONS` but this works too

benthecarman (approved):

Issue #569, opened 2023-09-29
Enumerate all possible payment failure states


futurepaul:
For design we want to be sure we know all the payment failure states the wallet can tell us about, and what actions are available to the user (and any "why" we can tell the user)

miawhite21:
I think we might have done this a while ago, here's what I have so far on the designs:
[Link to the designs](https://www.figma.com/file/jv72zrIZkNPz70qMsau0K9/Mutiny?type=design&node-id=2932%3A75955&mode=design&t=w0Vx5qsmtHvQGDCV-1)

1. **Payment route not found** (Solution: Retry later, request smaller invoices, or ask for a Bitcoin address.)
2. **Payment timed out** (Solution: Things to try to make your payment go through: Make sure your connection is good, Make sure the receiver’s connection is good , Make sure the receiver has their phone open, Maybe the LSP doesn’t have enough liquidity as the moment.)
3. **The Lightning invoice has expired.** (Solution: Ask the receiver for a new one.)
4. **Low balance - whether both, onchain, or LN** (Solution: load up balance, ask for smaller amount, or enter a smaller amount in case the user is the one setting the amount or use the other balance if it's just one that is low and you entered an unified invoice or swap some funds from one balance to the other). The solution for this one depends where you are on the flow when sending.
5. **Format not recognized** (Solution: Make sure you entered a valid Bitcoin address or invoice.)
6. **Wrong network** (Solution: make sure you enter an invoice, address for mainnet)

PR #677, opened 2023-11-08
Better payment timed out messaging

futurepaul:
we can handle other specific failures as well but this is a good start I think

Screenshot 2023-11-08 at 4 59 51 PM

Part of #569


src/routes/Send.tsx
```diff
@@ -635,15 +660,12 @@ export default function Send() {
...
                             <Match when={sentDetails()?.failure_reason}>
-                                <MegaEx />
-                                <h1 class="mb-2 mt-4 w-full text-center text-2xl font-semibold md:text-3xl">
-                                    {sentDetails()?.amount
-                                        ? source() === "onchain"
-                                            ? i18n.t("send.payment_initiated")
-                                            : i18n.t("send.payment_sent")
-                                        : sentDetails()?.failure_reason}
-                                </h1>
-                                {/*TODO: add failure hint logic for different failure conditions*/}
```
benthecarman, line 646: looks like you removed the on-chain success, but i dont know if this would ever be hit inside of the current `Match`. Also I see the same code below anyways

futurepaul: yeah my theory is that was just bad copy / pasting

src/logic/errorDispatch.ts
```diff
@@ -1,10 +1,12 @@
 // IMPORTANT: this should match 1:1 with the MutinyJsError enum in mutiny-wasm
 // If we can handle all of these, we can handle all the errors that Mutiny can throw
 
-// WARNING: autogenerated code, generated by errorsToTs.cjs
+// WARNING: autogenerated code, generated by calling:
+// `node scripts/errorsToTs.cjs /path/to/mutiny-wasm/src/error.rs`
```
benthecarman, line 5: Is there a way we could have CI check that this is up-to-date?

src/routes/Send.tsx
```diff
@@ -635,15 +660,12 @@ export default function Send() {
...
+                                <Failure
+                                    reason={
+                                        sentDetails()?.failure_reason ||
+                                        "Payment failed for an unknown reason"
```
benthecarman, line 666: nit: this `|| string` is redundant because we are checking it is defined in the `Match` if we can remove it it'd be better, if typescript yells at you fine to keep then, safer than putting a `!`

benthecarman (approved):
otherwise lgtm

src/routes/Send.tsx
futurepaul: yeah solidjs can't do type narrowing for these `if`-style components

src/logic/errorDispatch.ts
futurepaul: ideally this would be generated by wasm-bindgen (something like a string enum: https://github.com/rustwasm/wasm-bindgen/issues/3057) ... anything else is gonna be hacky

PR #678, opened 2023-11-09
Experimental websocket control channel

vfiano:
A work-in-progress experiment allowing another device to generate invoices.

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -10,6 +10,8 @@ import { Capacitor } from "@capacitor/core";
 import { SetupErrorDisplay, Toaster } from "~/components";
 
 import { useMegaStore } from "./state/megaStore";
+import {fiatToSats} from "~/utils";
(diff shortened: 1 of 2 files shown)


benthecarman:
This is certainly an interesting idea!

I don't see why you'd use this over nwc though?

futurepaul:
@vfiano this is a fun experiment but I'm going to close because I don't know why we'd merge this into the actual project. perhaps open an issue outlining your thinking and linking to this branch? thanks!

vfiano:
Hey guys! Sorry, I was hacking around last week and didn't mean for it to make work for you.

> I don't see why you'd use this over nwc though?

You're absolutely right, and the answer is because I didn't know about `nwc` until I read this comment. But now that I do I have more hacking to do.

benthecarman:
@vfiano we have nwc support in mutiny-node for paying invoices, happy to add support for creating invoices if you need it

PR #680, opened 2023-11-09
Min channel size to 100k

TonyGiorgio:


futurepaul (approved):
looks like you got all of them. def would be good if this was based on a constant and if I got rid of the `if bitcoin` logic, that's not helping anybody, but can do that later

PR #687, opened 2023-11-10
Check session storage is available

benthecarman:
testing to see if ios has this available, seems like it is working


benthecarman:
Did not find anything with this :(

PR #688, opened 2023-11-10
Another double init check

benthecarman:
Don't think this will fix any issues, before calling `setup` we check that state.mutinywallet is undefined but it can't hurt


futurepaul:
yeah I think it makes sense to defend inside here... refactored a bit

also I added defense to close #689

src/logic/mutinyWalletSetup.ts
```diff
@@ -187,7 +187,16 @@ export async function doubleInitDefense() {
...
     // Actually intialize the WASM, this should be the first thing that requires the WASM blob to be downloaded
-    await initMutinyWallet();
+
+    // If WASM is already initialized, don't init twice
+    try {
+        const _sats_the_standard = MutinyWallet.convert_btc_to_sats(1);
+        console.warn("MutinyWallet WASM already initialized, skipping init");
+        return;
+    } catch (e) {
+        console.warn("MutinyWallet WASM about to be initialized");
```
benthecarman, line 197: should make this a `debug`

futurepaul (approved):
we will triumph over the double inits!

PR #696, opened 2023-11-11
Handle iOS uris

benthecarman:
Not sure if we already have the handling for what happens after these


futurepaul:
afaik we don't handle any of these right now, so not sure it's correct to launch in response to them?

PR #702, opened 2023-11-14
Use router

futurepaul:
this PR gets rid of solid-start and instead just relies on solid-router and solid-meta 

https://github.com/solidjs/solid-router
https://github.com/solidjs/solid-meta


futurepaul:
all tests pass because I marked the restore test as `@slow` and filtered it out.

e2e/routes.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+        // Currency
+        await checkRoute(page, "/settings/currency", "Currency", checklist);
+        await page.goBack();
+
+        // Servers
+        await checkRoute(page, "/settings/servers", "Servers", checklist);
+        await page.goBack();
+
+        // Connections
+        await checkRoute(page, "/settings/connections", "Wallet Connections", checklist);
```
TonyGiorgio, line 80: Can you test the wallet connections page with query parameters too? Like you did with the gift one below.

TonyGiorgio (changes requested):

TonyGiorgio:
Chrome and Safari look good to me. Still want to test android/ios pwa and then the native apps.

e2e/routes.spec.ts
futurepaul: added a test for wallet connections with params

TonyGiorgio:
Tests failing

TonyGiorgio (commented):
Tested pretty much every page on iOS and Android native & pwa. Just need the tests to pass reliably and I'd say it's good to merge whenever.

futurepaul:
the one that failed was the send / receive test which hasn't failed before... could re-run again, but I feel pretty good about it. ran really quick locally

TonyGiorgio (approved):
Ran it 5 times and it passed. Nice work!

PR #709, opened 2023-11-15
Playwright CI improvements

benthecarman:
Makes it so we will only run on PRs (no reason to run it on merge)

Also makes it so only one will run at a time per PR, this way we'll have a higher likelihood of things not getting borked

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,9 +1,11 @@
 name: Playwright Tests
+
+# Only run one at a time per branch
+concurrency:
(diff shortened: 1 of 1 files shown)


TonyGiorgio (approved):

.github/workflows/playwright.yml
```diff
@@ -1,9 +1,11 @@
 name: Playwright Tests
+
+# Only run one at a time pre branch
```
futurepaul, line 3: sry but per branch?

futurepaul (approved):
nice

PR #714, opened 2023-11-15
don't refetch connections every sync

### LINKED ISSUE #708: NWC Dialog Closing
When you click on a NWC profile in the Wallet Connections page it will close quickly after. This happening when the page is refetching the list of nwc profiles.

futurepaul:
closes #708 

this is one of those "I don't know why it was working before" kind of bugs but seems solid now. basically we just refetch after edits, instead of constantly


benthecarman (approved):
works for me!

one downside is that with this you couldnt watch your budget go down live, but I dont think that ever worked before

futurepaul:
> one downside is that with this you couldnt watch your budget go down live, but I dont think that ever worked before

yeah we could do refetching inside a single one, just need to not be refetching this whole list every time

PR #716, opened 2023-11-16
sync price every minute and backoff if err

futurepaul:
separates price syncing from balance syncing, sets the price fetch interval to a minute, backs off if it gets errors, and tries to get price on boot even if we have zero balance (so the first receive can be priced in fiat)

seems to be working in my local testing. my apologies to coingecko for all the spam

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -50,6 +50,7 @@ export type MegaStore = [
         balance?: MutinyBalance;
         is_syncing?: boolean;
         last_sync?: number;
+        price_sync_backoff_multiple?: number;
         price: number;
         fiat: Currency;
         has_backed_up: boolean;
@@ -105,6 +106,7 @@ export const Provider: ParentComponent = (props) => {
         has_backed_up: localStorage.getItem("has_backed_up") === "true",
         balance: undefined as MutinyBalance | undefined,
         last_sync: undefined as number | undefined,
+        price_sync_backoff_multiple: 1,
         is_syncing: false,
         wallet_loading: true,
         setup_error: undefined as Error | undefined,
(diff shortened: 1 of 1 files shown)


benthecarman (changes requested):
you never reset `price_sync_backoff_multiple` back to `1` on successful fetches

src/state/megaStore.tsx
```diff
@@ -215,25 +217,17 @@ export const Provider: ParentComponent = (props) => {
                 // Get balance + price optimistically
                 const balance = await mutinyWallet.get_balance();
                 let price;
-                // only get price if balance is non-zero
-                if (
-                    balance.confirmed > 0 ||
-                    balance.unconfirmed > 0 ||
-                    balance.lightning > 0 ||
-                    balance.force_close > 0
-                ) {
```
benthecarman, line 224: this was to improve our startup speed, might be worth keeping

src/state/megaStore.tsx
```diff
@@ -459,6 +465,14 @@ export const Provider: ParentComponent = (props) => {
         setInterval(async () => {
             await actions.sync();
```
benthecarman, line 466: i think actions.sync() gets the price too, should do a sweep to see where we are fetching the price

src/state/megaStore.tsx
futurepaul: if we don't do it here, we have to put in logic to somehow fetch the price before the user's first receive. it's nice to speed up startup times but seems like it complicates the logic a lot

src/state/megaStore.tsx
futurepaul: this pr updates sync() to not do price anymore

futurepaul:
fixed the backoff reset if you want to re-review

benthecarman (approved):

TonyGiorgio:
Seems to work for me when I was getting some 403's

TonyGiorgio (approved):

PR #724, opened 2023-11-17
Show lightning address on send screen

benthecarman:
Before if you entered a lightning address the destination would be it lnurl encoded, this will fix for it so show as the normal lightning address.

Also adds the is hodl invoice check for zeuspay addresses

image

### DIFF
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -179,6 +179,7 @@ function DestinationShower(props: {
     invoice?: MutinyInvoice;
     nodePubkey?: string;
     lnurl?: string;
+    lightning_address?: string;
     clearAll: () => void;
 }) {
     return (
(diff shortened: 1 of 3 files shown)


src/routes/Send.tsx
```diff
@@ -192,7 +193,18 @@ function DestinationShower(props: {
             <Match when={props.nodePubkey && props.source === "lightning"}>
                 <StringShower text={props.nodePubkey || ""} />
             </Match>
-            <Match when={props.lnurl && props.source === "lightning"}>
+            <Match
+                when={props.lightning_address && props.source === "lightning"}
+            >
+                <StringShower text={props.lightning_address || ""} />
```
futurepaul, line 199: string shower is meant for longer strings, it chokes on longish lightning addresses

Screenshot 2023-11-17 at 1 53 51 PM

I can fix if you'd like

futurepaul (changes requested):
nice ux improvement these lnurl strings always bothered me. just need to fix the string shower thing

src/routes/Send.tsx
benthecarman: yeah can do you do that

futurepaul: didn't end up with a truly beautiful solution but it works... seems like the long term perfect solution to a middle ellipses is some container-measuring shenanigans

benthecarman: cool update looks good to me

futurepaul (approved):

PR #725, opened 2023-11-17
test caching with service worker

futurepaul:
now that our bundling process is simpler (#702) hopefully we can do smooth upgrades with service worker caching

basic idea is the service worker creates a cache of the assets, then when the sw notices there are new versions of those assets, it loads them into its and then prompts the user to reload the app (after which, the assets will be served from the sw cache).

if the user does nothing, the new versions will be used at next load. (no way to postpone an update indefinitely from my research).

the big items are, for example:
`dist/public/assets/index-49c6ccd6.js`
and
`dist/public/assets/mutiny_wasm_bg-4f199409.wasm`

the problem we had before was there were a lot more large "chunks" and it was possible to end up with some chunks updated, some not. now all the frontend logic is inside the big index.js chunk. I can't think of a way that the index chunk could be updated out of sync with the wasm chunk, but **that's the important thing to verify**!

want to push updates to this branch and see how the upgrade process goes

other stuff that could potentially go wrong:
* I added an app id, which https://www.pwabuilder.com/ recommended ... it seems like the right thing to do, but there's maybe a 
* cloudflare's cache has been wonky lately. cloudflare cache + sw cache = idk what
* in vite.config.ts there's a "devOptions" field. if you enable that, the service worker will run locally. due to the weirdness of caching when you're jumping between branches and stuff, enable this locally at your own risk. browsers have some debug options for clearing out / manually updating service workers if stuff goes wrong


vite.config.ts
```diff
@@ -1,22 +1,30 @@
...
-    registerType: "autoUpdate",
+    registerType: "prompt",
     devOptions: {
         enabled: false
     },
+    workbox: {
+        navigateFallback: "/index.html",
+        globPatterns: ["**/*.{js,css,html,svg,png,gif,wasm}"],
+        // mutiny_wasm is 10mb, so we'll do 12mb to be safe
+        maximumFileSizeToCacheInBytes: 12 * 1024 * 1024
```
benthecarman, line 26: Should make this like 50 to be safe

src/routes/Main.tsx
```diff
@@ -93,6 +93,7 @@ export function Main() {
                         </Suspense>
                     </VStack>
                 </Card>
+                <pre>hey tony</pre>
```
TonyGiorgio, line 96: remove this

TonyGiorgio (changes requested):

futurepaul:
cloudflare hiccup on the last push but it was just a rebase collapsing commits so should be all good

src/components/Reload.tsx
```diff
@@ -17,27 +23,47 @@ export function ReloadPrompt() {
...
+            <div class="grid grid-cols-[auto_minmax(0,_1fr)_auto] gap-4 rounded-xl bg-neutral-950/50 p-4">
+                <div class="self-center">
+                    <img src={refresh} alt="refresh" class="h-8 w-8" />
+                </div>
+                <div class="flex flex-row justify-between gap-4 max-md:items-center">
+                    <div class="flex flex-col">
+                        <SmallHeader>Mutiny Update</SmallHeader>
+                        <p class="text-base font-light max-md:hidden">
+                            New version of Mutiny has been cached, reload to
+                            start using it.
```
TonyGiorgio, line 45: Stick these in the translator?

futurepaul: good catch

TonyGiorgio:
Not sure why you're adding other things to this PR but let's get this caching stuff merged asap so we can test it out on signet this week.

futurepaul:
thinking about service worker stuff got me thinking about manifest stuff got me curious about why ios is blank white during launch... didn't solve it, I can revert these

PR #726, opened 2023-11-20
Free gifting on thanksgiving

benthecarman:
Can test by changing your system clock, works for me when changing to this thursday!


futurepaul (approved):
great job

PR #727, opened 2023-11-20
check for subscribed on startup

futurepaul:
this was pretty easy mostly just deleting code

(btw reason for duplication between checkForSubscription and the logic in setup is there's no mutiny_wallet object yet for the action version to use)

### DIFF
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -133,20 +133,16 @@ export const Provider: ParentComponent = (props) => {
         async checkForSubscription(justPaid?: boolean): Promise<void> {
             try {
                 const timestamp = await state.mutiny_wallet?.check_subscribed();
-                console.log("timestamp:", timestamp);
-                if (timestamp) {
-                    localStorage.setItem(
-                        "subscription_timestamp",
-                        timestamp?.toString()
-                    );
+
+                // Check that timestamp is a number
+                if (timestamp && !isNaN(Number(timestamp))) {
                     setState({ subscription_timestamp: Number(timestamp) });
-                }
-            } catch (e) {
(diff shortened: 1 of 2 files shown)


src/state/megaStore.tsx
```diff
@@ -192,30 +191,18 @@ export const Provider: ParentComponent = (props) => {
...
-                        );
-                    } catch (e) {
-                        console.error(e);
+                // Check if we're subscribed and update the timestamp
+                try {
+                    const timestamp = await mutinyWallet?.check_subscribed();
+
+                    // Check that timestamp is a number
+                    if (!timestamp || isNaN(Number(timestamp))) {
+                        throw new Error("Timestamp is not a number");
```
benthecarman, line 200: Would be nice to not throw the error when `timestamp` is undefined so we dont get the scary error log on every new wallet created

src/state/megaStore.tsx
```diff
@@ -133,14 +133,13 @@ export const Provider: ParentComponent = (props) => {
...
-                if (timestamp) {
-                    localStorage.setItem(
-                        "subscription_timestamp",
-                        timestamp?.toString()
-                    );
-                    setState({ subscription_timestamp: Number(timestamp) });
+
+                // Check that timestamp is a number
+                if (!timestamp || isNaN(Number(timestamp))) {
+                    throw new Error("Timestamp is not a number");
```
benthecarman, line 139: same here

src/state/megaStore.tsx
futurepaul: fixed, can you double check my booleans?

benthecarman (approved):

PR #730, opened 2023-11-20
Countdown timer for multi device

benthecarman:
requires https://github.com/MutinyWallet/mutiny-node/pull/868


src/components/SetupErrorDisplay.tsx
```diff
@@ -127,6 +164,9 @@ export function SetupErrorDisplay(props: { initialError: Error }) {
                                 "error.on_boot.already_running.description"
                             )}
                         </NiceP>
+                        <p class="rounded-xl bg-white/10 p-4 font-mono">
+                            Retry again in {lockSeconds()} seconds
```
TonyGiorgio, line 168: This should be in i18n

TonyGiorgio (changes requested):
This screws up on the `Error: Existing tab detected, aborting setup` screen, refreshes every 1 second.

TonyGiorgio (approved):
works for me

futurepaul:
added a commit so password works. will pass the password to the setup error component (or undefined if none). still need to re-enter your password after reload but that doesn't feel like a big deal

TonyGiorgio (approved):
Tested a few cases, works for me!

PR #737, opened 2023-11-24
Reload page on Wasm error

benthecarman:
When we get this error you just need to refresh the page, so we can just do it automatically.

One potential issue is that if refreshing doesn't fix, it'll be in a refresh loop

image

### DIFF
--- a/src/components/ErrorDisplay.tsx
+++ b/src/components/ErrorDisplay.tsx
@@ -38,6 +38,9 @@ export function ErrorDisplay(props: { error: Error }) {
                         {i18n.t("error.general.support_link")}
                     </ExternalLink>
                 </NiceP>
+                <Button onClick={() => window.location.reload()}>
+                    {i18n.t("error.reload")}
+                </Button>
                 <NiceP>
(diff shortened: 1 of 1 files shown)


benalleng:
Is it worth giving the user a button to refresh the page here? On PWA's since there is no drag to refresh the loop could be avoided if we required the user to refresh, admittedly still worse UX that automatically refreshing

benthecarman:
Probably is worth adding a refresh button, a lot of the times we have random errors that are fixed by a refresh

futurepaul:
I haven't seen this error lately. But how about we just add the reload button to `ErrorDisplay.tsx` (I think that's what is being shown here, not `SetupErrorDisplay.tsx`)

```
<Button onClick={() => window.location.reload()}>
    {i18n.t("error.reload")}
</Button>
```

futurepaul (approved):
@benthecarman I changed this PR to just add the reload button to the other error page. if you ack I can merge

benthecarman (commented):
ACK

PR #746, opened 2023-11-29
Federations

TonyGiorgio:
Depends on https://github.com/MutinyWallet/mutiny-node/pull/867


src/i18n/en/translations.ts
```diff
@@ -520,6 +520,21 @@ export default {
...
+            federation_code_required: "Federation code can't be blank",
+            federation_added_success: "Federation added successfully",
+            federation_remove_confirm:
+                "Are you sure you want to remove this federation? Make sure any funds you have are transferred to your lightning balance or another wallet first.",
+            add: "Add",
+            remove: "Remove",
+            expires: "Expires",
+            federation_id: "Federation ID",
+            description:
+                "Mutiny has experimental support for the Fedimint protocol. You'll need a federation invite code to use this feature. Store funds in a federation at your own risk!",
```
TonyGiorgio, line 535:
```suggestion
                "Mutiny has experimental support for the Fedimint protocol. You'll need a federation invite code to use this feature. These funds are currently not backed up remotely. Store funds in a federation at your own risk!",
```

futurepaul (approved):
lgtm / works on my machine. tested on chrome and in safari on ios and as ios app on device

src/routes/Gift.tsx
```diff
@@ -67,7 +67,9 @@ function InboundWarning() {
         const network = state.mutiny_wallet?.get_network() as Network;
 
         const threshold = network === "bitcoin" ? 100000 : 10000;
-        const balance = state.balance?.lightning || 0n;
+        const balance =
+            (state.balance?.lightning || 0n) +
+            (state.balance?.federation || 0n);
```
benthecarman, line 72: we are doing this a bunch everywhere, probably a good indication we should add in mutiny-core something like `can_pay_lightning(amt)` that checks both federation and lightning balance. Good for a follow up though

src/routes/settings/ManageFederations.tsx
```diff
@@ -0,0 +1,219 @@
...
+                >
+                    {(field, props) => (
+                        <TextField
+                            {...props}
+                            {...field}
+                            error={field.error}
+                            label={i18n.t(
+                                "settings.manage_federations.federation_code_label"
+                            )}
+                            placeholder="fedi1..."
```
benthecarman, line 96: `i` isn't a valid bech32 character it actually starts with `fed11`

src/i18n/en/translations.ts
```diff
@@ -533,6 +533,21 @@ export default {
...
+            federation_code_required: "Federation code can't be blank",
+            federation_added_success: "Federation added successfully",
+            federation_remove_confirm:
+                "Are you sure you want to remove this federation? Make sure any funds you have are transferred to your lightning balance or another wallet first.",
+            add: "Add",
+            remove: "Remove",
+            expires: "Expires",
+            federation_id: "Federation ID",
+            description:
+                "Mutiny has experimental support for the Fedimint protocol. You'll need a federation invite code to use this feature. These funds are currently not backed up remotely. Store funds in a federation at your own risk!",
```
benthecarman, line 548: Could note something about rug-pull risk here too, fine for now though

src/routes/settings/ManageFederations.tsx
```diff
@@ -0,0 +1,219 @@
...
+                </Show>
+                <Show when={props.fed.welcome_message}>
+                    <p>{props.fed.welcome_message}</p>
+                </Show>
+                <Show when={props.fed.federation_expiry_timestamp}>
+                    <KeyValue
+                        key={i18n.t("settings.manage_federations.expires")}
+                    >
+                        <time>
+                            {timeAgo(props.fed.federation_expiry_timestamp)}
```
benthecarman, line 160: `timeAgo` isn't the best name for this function now since it is returning a future time, but fine for now i guess

benthecarman: Would be good to display a warning or something if the expiration is in the past though

benthecarman (changes requested):

benthecarman (approved):

PR #747, opened 2023-11-30
Nostr Wallet Auth support

futurepaul:
- [x] needs https://github.com/MutinyWallet/mutiny-node/pull/815
- [ ] needs https://github.com/futurepaul/zapple-pay-frontend/pull/13

hard part here was refactoring the nwc connect form to handle all the different scenarios it might find itself in (NWCBudgetEditor -> NWCEditor)

there's a FIXME in here because when you go to edit budget it's supposed to pull up the profile from our contacts but I'm getting undefined for some reason. I don't think it's a blocker, just a nice to have (the profile name is the name of the connection so it shouldn't be too ambiguous.


src/components/NWCEditor.tsx
```diff
@@ -0,0 +1,522 @@
...
+            return contact;
+        } catch (e) {
+            console.error(e);
+            return undefined;
+        }
+    };
+
+    const [contact] = createResource(
+        // props.initialProfile?.label,
+        // undefined,
```
TonyGiorgio, line 212: can remove these?

src/components/NWCEditor.tsx
```diff
@@ -0,0 +1,522 @@
...
+type FormMode = "createnwa" | "createnwc" | "editnwc";
+type BudgetMode = "fixed" | "editable";
+
+function parseNWA(nwaString?: string) {
+    if (!nwaString) return undefined;
+    const nwa = decodeURI(nwaString);
+    if (nwa) {
+        // Examples:
+        // nostr+walletauth://a957bc527d4b7cea5134308412719fa675671ed38eb313adcf89b96c6982480e?relay=wss%3A%2F%2Frelay.damus.io%2F&secret=a2522b4c6d6ae729&required_commands=pay_invoice&budget=21%2Fday&identity=04c915daefee38317fa734444acee390a8269fe5810b2241e5e6dd343dfbecc9
+        // nostr+walletauth://c95053d4df36104c7d1a007c4c46fd195920d268527c21d5474b1842ef2e4170?relay=wss%3A%2F%2Frelay.damus.io%2F&secret=39133ed58614e67b&required_commands=pay_invoice&budget=21%2Fday&identity=0d6c8388dcb049b8dd4fc8d3d8c3bb93de3da90ba828e4f09c8ad0f346488a33
```
benthecarman, line 55: these 2 examples are exactly the same besides different keys. not a huge deal but don't see the point of having both

src/components/NWCEditor.tsx
```diff
@@ -0,0 +1,517 @@
...
+            return undefined;
+        }
+    };
+
+    const [profile] = createResource(
+        props.initialProfileIndex,
+        nwcProfileFetcher
+    );
+
+    // TODO: this should get the contact so we can get the image, but not getting a contact tagged on the nwc right now
```
benthecarman, line 192: it looks like this is done

TonyGiorgio (approved):

benthecarman:
otherwise lgtm

src/components/NWCEditor.tsx
futurepaul: not getting the image url in the contact response, we can revisit

PR #748, opened 2023-12-01
Latest LSPS integration

johncantrell97:
PR for testing with my fork of mutiny-wasm.  Though this does have the goal of making it into the wallet as an off-by-default but admin option a user can manually enable.

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,10 +4,10 @@ dev:
     pnpm run dev
 
 local:
-    pnpm install && pnpm link --global "@mutinywallet/mutiny-wasm"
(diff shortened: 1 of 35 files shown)


johncantrell97:
yeah once that PR gets merged it doesn't need to use my fork of mutiny-node

futurepaul:
Okay I think what I'll do is make my own PR that adds multi-lsp to the admin page so we can beta test without exposing this broadly at first.

futurepaul:
closing in favor of #767 thanks for all your work on this!

PR #763, opened 2023-12-08
add lnurlauth scanning

futurepaul:
removes the placeholder page for lnurlauth in settings. you can scan lnurls now! also a little easter egg


benthecarman:
this wont work for me

futurepaul:
I think this is the browser protecting you from malicious paste requests. On a fresh chrome incognito I have to click a couple times before I get the permission prompt. On mobile safari I have to click the little "paste" prompt that pops up above the button. Should be the same behavior with pasting an invoice.

benthecarman (approved):
Worked on my phone

PR #766, opened 2023-12-12
Fix nginx for other routes

benthecarman:
Previously if you went to any route that wasn't `/` on a self hosted instance it would fail unless you originally navigated from the `/` route. This fixes so those routes will fallback to the `index.html` so it can get the correct page


futurepaul (approved):
haven't tested but makes sense to me. all routes that aren't .css, image, .js or .wasm assets need to be directed to index.html

PR #771, opened 2023-12-15
switch Contact to TagItem, handle edit

futurepaul:
requires https://github.com/MutinyWallet/mutiny-node/pull/898

also this change allows us to do contact editing so I added a small amount of that (you can also add a lightning address when creating... good test is to add faucet as a contact: `refund@lnurl-staging.mutinywallet.com`)

didn't go crazy with other possibilities because contacts are going to shift around a lot in the new design


TonyGiorgio (approved):
Full testing worked

PR #775, opened 2023-12-20
new send

futurepaul:
this is a bit of a monster but send screen redesign led to amounteditable redesign led to changes everywhere that needs editable amounts. so far I'm actually net positive on deleting code but I'm sure there are some regressions I'll need to shore up.

TODO:
- [x] handle which contact you have selected (needs #771)
- [x] make sure sends look good when not sending to a contact (on-chain, raw invoice)
- [x] make sure we're displaying fees in all the places we should
  - [x] on-chain send
  - [x] swap
- [x] make sure we don't have redundant / duplicate warnings
- [x] "max amount" button
- [x] test on android and ios, both in app and in browser, because we're reliant on the device keyboard opening now
- [x] when testing on devices make sure we got the safe areas correct (seems to be working on ios browser so far)
- [x] rebase to nice commits

Follwup todo:
- add contact from search for receives
- toggle private vs public zap for sends


benthecarman:
image

The warning message gets cut off by my keyboard 

Also after you dismiss the keyboard to get it back you have to click the number which switches it to USD

benthecarman:
Create contact should have a npub field

benthecarman:
I created a contact and when it went to the send screen it said my balance was 0. After I left and went back it had the correct balance

benthecarman:
image

imagea

Should unify these so they look the same (the contact picture)

TonyGiorgio:
Screenshot 2024-01-03 at 8 31 31 PM

Don't allow someone to attempt to change the invoice amount, remove the max buttons

TonyGiorgio:
Screenshot 2024-01-03 at 8 59 24 PM

Can you see if you can make the `->` arrow do the confirmation? That's the first thing I was trying to click.

TonyGiorgio:
Screenshot 2024-01-03 at 9 22 32 PM

Manually pasting this invoice in, instead of using the `Paste` button resultls in an invalid invoice warning. Meanwhile clicking the paste button works. 

```
lnbc100n1pjevfqkpp5608hnh4k5c0t2puuxxvzgp8suvftj9xtd5a3pvsya7pslx6lg90sdqqcqzzsxqrrs0sp556k66c67ych9f0ratcwpchzpa40hehjqatja5qv5wujw7c92pyuq9qyyssqeztt58kmcglhlcqewq36smzzgdlzuyhrrammqagsz2zxvgfnjffs430yjklc8pw6a9lzchtd22ctfdv9ptrkjhjvdzdu5ltjldtkucsppapn6d
```

TonyGiorgio:
Screenshot 2024-01-03 at 9 32 56 PM

The emulator isn't popping up the keyboard, is that expected?

TonyGiorgio:
I've tested this on multiple devices where I have already sync my contacts once and it does not show my existing contacts at all when I try out this version. I have to resync on each device first.

TonyGiorgio:
Screenshot 2024-01-03 at 9 46 55 PM

Should just default to our empty profile pic version if the image fails to load.

TonyGiorgio:
The send screen is cut off on iOS 

Screenshot 2024-01-03 at 9 48 05 PM

futurepaul:
alright I addressed all of your comments above, here's where we're at now:

### done 
- [x] remove gift button
- [x] add cta for import contacts from nostr on the "add contact" screen
- [x] "I created a contact and when it went to the send screen it said my balance was 0. After I left and went back it had the correct balance"
- [x] unify "contact picture" when there's no picture
- [x] hide max button if there's an amount already
- [x] make android arrow do the "confirm send" action (also works for receive)
- [x] ios "done" button removes focus, so now it won't switch when you tap the amount to refocus
- [x] manually pasting invoice, instead of `paste` button results in invalid invoice
- [x] can I hide the empty profile pic if image fails to load?
- [x] if I send too fast it goes to "success" but also sent zero
- [x] change the receive screen to use the same layout as send
- [x] fix that focus box showing up on the ? button
- [x] send screen is cut off?

### won't fix
- [ ] warning message gets cut off by keyboard (sry but it's just too much hassle fighting css + native keyboards + viewport shifts... it works pretty good on android at least. on ios it doesn't seem like the viewport is changing in size)

### blocked
- [ ] contact should have npub field (@benthecarman said he could give me npub in / out instead of me needing to do conversions to / from hex everywhere)

### can't repro
- [ ] I have to resync on each device first. (@TonyGiorgio I tried to do this a couple times where I started with an old version, then switched to the new version. my contacts came across each time)

futurepaul:
oh, forgot to ask: how should I be calculating "max" for send nowadays?

TonyGiorgio:
> oh, forgot to ask: how should I be calculating "max" for send nowadays?

For LN sends? Probably `max(ln_balance, fedimint_balance)`

benthecarman:
decimals for usd amounts get screwy, showed irl

TonyGiorgio:
Wow, this has come a long way, nice job! 

A few things: 
- Good catch Ben, entering an amount in the fiat mode is a bit funky
- In android on the received screen, I love that the "->" keyboard button does the `continue` action. However, in the "What's this for?" dialog, the "->" doesn't do the same. A small nit, but a nice to have if you have time or create a follow up issue. 
- RE: "warning message gets cut off by keyboard", maybe you can just cut down on the text here? For instance, since we show the fee on the next page, maybe we just say "first receive incurs a cost" and leave it at that. Next screen gives a similar warning and shows the exact fee. It might still be a bit cut off, but I bet we can go from 3 lines of text to just one line here. 

Only the fiat mode thing would be a blocker IMO. Other than that, would LOVE to get this deployed by thursday for demo's.

futurepaul:
okay latest push:

- [ ] rebased
- [ ] "continue" / enter in the "What's this for?" field should submit on receive and send
- [ ] shortened the amount warning
- [ ] fixed typing in fiat mode (was overwriting my bespoke state with some dumb reactivity)
- [ ] fixed a bug where the search box loses focus (due to profile images not loading causing an error... would eventually like to get profile images cached in some nice way so it's not a network fiesta)
- [ ] lightning address is required for a new contact now (still some improving to do with handling what happens if they put in a bad ln address)

benthecarman (approved):

TonyGiorgio (approved):

PR #779, opened 2023-12-21
Drop MutinyWallet before restore

TonyGiorgio:
With gluedb, we need to drop all instances of it before trying to init it again. Since we're doing associated functions here, we need to init gluedb out of band since we don't rely on `MutinyWallet` anymore after we stop it. This makes sure we stop mutiny wallet and then drop it then do DB operations.


futurepaul (approved):
works for me!

PR #781, opened 2023-12-24
Remove redshift

TonyGiorgio:
Redshift will be removed in the next release candidate of mutiny-node: https://github.com/MutinyWallet/mutiny-node/pull/913


futurepaul (approved):
rip

Issue #639, opened 2023-10-28
Add COP ticker


richardelima:
Please add support for colombian pesos. COP
Thanks in Advance

benalleng:
Unfortunately, at this time the currencies we can add support for are limited by the coingecko API
https://api.coingecko.com/api/v3/simple/supported_vs_currencies

benthecarman:
Our latest release added support for COP currency

PR #797, opened 2024-01-05
Remove no-longer supported currencies + add new ones

benthecarman:
the new price api doesn't support these, these aren't huge countries so i think it is fine to remove for now, if someone complains we add another provider to the price api to bring them back

COP is now supported which a user requested: #639

test with: https://github.com/MutinyWallet/mutiny-node/pull/936

### DIFF
--- a/src/utils/currencies.ts
+++ b/src/utils/currencies.ts
@@ -163,20 +163,9 @@ export const FIAT_OPTIONS: Currency[] = [
         maxFractionalDigits: 0
     },
     {
-        label: "Kuwaiti Dinar KWD",
-        value: "KWD",
-        maxFractionalDigits: 3
-    },
-    {
         label: "Sri Lankan Rupee LKR",
         value: "LKR",
         hasSymbol: "Rs",
-        maxFractionalDigits: 2
-    },
-    {
-        label: "Myanmar Kyat MMK",
-        value: "MMK",
-        hasSymbol: "Ks",
         maxFractionalDigits: 2
     },
     {
@@ -273,12 +262,6 @@ export const FIAT_OPTIONS: Currency[] = [
         label: "Ukrainian Hryvnia UAH",
         value: "UAH",
         hasSymbol: "₴",
-        maxFractionalDigits: 2
-    },
-    {
-        label: "Venezuelan Bolívar VEF",
-        value: "VEF",
-        hasSymbol: "Bs",
         maxFractionalDigits: 2
     },
     {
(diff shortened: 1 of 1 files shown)


TonyGiorgio:
Are there any more we can add? Just add them all.

benthecarman:
> Are there any more we can add? Just add them all.

yeah probably, will go through and try and add them

benthecarman:
Okay added all the new currencies 59 in total!

src/utils/currencies.ts
```diff
@@ -287,6 +594,36 @@ export const FIAT_OPTIONS: Currency[] = [
         hasSymbol: "₫",
         maxFractionalDigits: 0
     },
+    {
+        label: "Central African CFA Franc BEAC XAF",
+        value: "XAF",
+        hasSymbol: "CFA Franc BEAC",
+        maxFractionalDigits: 2
+    },
```
benalleng, line 602: For the African Franc I think they should be denoted like this.
```suggestion
    {
        label: "Central African CFA Franc XAF",
        value: "XAF",
        hasSymbol: "F.CFA",
        maxFractionalDigits: 2
    },
```

src/utils/currencies.ts
```diff
@@ -287,6 +594,36 @@ export const FIAT_OPTIONS: Currency[] = [
...
+        value: "XAU",
+        hasSymbol: "Oz",
+        maxFractionalDigits: 6
+    },
+    {
+        label: "CFA Franc XOF",
+        value: "XOF",
+        hasSymbol: "Franc",
+        maxFractionalDigits: 2
+    },
```
benalleng, line 620: And for the West African Franc
```suggestion
    {
        label: "West African CFA Franc XOF",
        value: "XOF",
        hasSymbol: "F.CFA",
        maxFractionalDigits: 2
    },
```

benalleng (commented):
Haven't had a chance to test yet, but I caught 2 values i think could be labeled better. Going to double check with someone who I think probably knows better

benthecarman:
I just got all the names from xe.com, not really sure of a better source

src/utils/currencies.ts
0tuedon: sad to say it does have a single character sure, but I think either F.CFA or FCFA is fine, but I will prefer the latter @benalleng 
cc: @benthecarman

futurepaul (approved):
lgtm once we get the new version of mutiny-wasm out

TonyGiorgio:
Is this good now?

benthecarman:
Should be good for merge

PR #798, opened 2024-01-05
Skip invoice paid twice error in NWC

benthecarman:
We handle this for auto approvals in the backend but can't for manual. If we've already paid an invoice no real reason to show an error, they've already paid it so only way forward is to skip and go on so this just changes to skip those errors

### DIFF
--- a/src/components/PendingNwc.tsx
+++ b/src/components/PendingNwc.tsx
@@ -78,8 +78,20 @@ export function PendingNwc() {
             await state.mutiny_wallet?.approve_invoice(item.id);
             await vibrateSuccess();
         } catch (e) {
-            setError(eify(e));
-            console.error(e);
+            const err = eify(e);
+            // If we've already paid this invoice, just ignore the error
+            // we just want to remove it from the list and continue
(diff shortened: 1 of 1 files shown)


TonyGiorgio (approved):
concept ack

src/components/PendingNwc.tsx
```diff
@@ -78,8 +78,19 @@ export function PendingNwc() {
             await state.mutiny_wallet?.approve_invoice(item.id);
             await vibrateSuccess();
         } catch (e) {
-            setError(eify(e));
-            console.error(e);
+            // If we've already paid this invoice, just ignore the error
+            // we just want to remove it from the list and continue
+            if (e === "An invoice must not get payed twice.") {
```
futurepaul, line 83: it's usually `e.message`, not the `e` object itself. is there a way we can test this?

futurepaul (changes requested):

src/components/PendingNwc.tsx
benthecarman: hard to test, this is a pretty uncommon edge case

futurepaul (approved):

PR #806, opened 2024-01-08
Truncation underline bridge

### LINKED ISSUE #789: Underline does not cover whole txid
image

benalleng:
Closes #789 

Before:
image

After:
image

### DIFF
--- a/src/components/ActivityDetailsModal.tsx
+++ b/src/components/ActivityDetailsModal.tsx
@@ -405,6 +405,7 @@ function OnchainDetails(props: {
                                         />
                                     </svg>
                                 </div>
+                                <hr class="truncate-bridge h-px w-[80%] border-m-grey-400" />
                             </a>
                         </div>
                         <button
--- a/src/root.css
+++ b/src/root.css
@@ -80,3 +80,19 @@ select {
 strong {
     @apply font-semibold text-m-red;
 }
+
+/* different default spacing required some broswer specific styles for a link underline */
+.truncate-bridge {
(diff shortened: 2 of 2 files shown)


benthecarman (approved):

futurepaul:
doesn't look quite right on mac chrome. I haven't messed with this in a while but is there a reason we don't just use a `<span>…</span>` here?

Screenshot 2024-01-10 at 12 20 27 PM

benalleng:
So a normal span doesn't work exactly as I want because the behavior of an overflow is to clip instead of delete characters so we end up with something like below when using `<span>…</span>` curious if you think you prefer this over trying to adjust the `<hr/>` for all browsers

image

benalleng:
I tried setting a default underline offset as it seems like there is a difference between firefox and chrome, I tested it myself on a few browsers but I haven't gotten a chance to test on mac

futurepaul:
it's worse on safari for some reason:

Screenshot 2024-01-10 at 9 47 05 PM

> So a normal span doesn't work exactly as I want because the behavior of an overflow is to clip instead of delete

ah, the pain is coming back to me lol

futurepaul:
> image

yeah maybe this can be fudged a bit with some margins or negative margins?

benalleng:
Ok i tried again with some adjustments, I ended up not using relative and positionioning and instead opted for a browser based negative margin, also tested on my girlfriend's iPhone and it looks like it worked! Let me know if you're cool with the browser based styling

PR #809, opened 2024-01-10
Deploy to prod

benthecarman:


futurepaul (approved):

benthecarman (commented):
tested new wallet and setting up a nwc, seems like it is all working.

PR #810, opened 2024-01-10
Bump to v0.5.3

benthecarman:


TonyGiorgio (approved):

src/routes/Send.tsx
```diff
@@ -567,7 +567,8 @@ export function Send() {
                     lnurlp()!,
                     amountSats(),
                     undefined, // zap_npub
-                    tags
+                    tags,
+                    whatForInput().trim()
```
TonyGiorgio, line 571: Hmm. Nack IMO. We should be distinguishing between what is public and what is private.

TonyGiorgio (changes requested):

src/routes/Send.tsx
benthecarman: Good point, we can bring this back when we have the switcher

futurepaul (approved):
lgtm will try to get the public / private toggle in tmrw

PR #811, opened 2024-01-10
add global contact search

futurepaul:
uses primal's search api to search everybody on nostr

when testing on signet you can search mutinynet and you get the refund address

clicking a search result will go to the send screen and add that user to your contacts

### DIFF
--- a/src/utils/fetchZaps.ts
+++ b/src/utils/fetchZaps.ts
@@ -28,7 +28,7 @@ type SimpleZapItem = {
     content?: string;
 };
 
-type NostrProfile = {
+export type NostrProfile = {
     id: string;
     pubkey: string;
     created_at: number;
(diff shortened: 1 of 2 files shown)


futurepaul:
https://global-search.mutiny-web.pages.dev/

src/utils/fetchZaps.ts
```diff
@@ -316,3 +316,60 @@ export const fetchNostrProfile: ResourceFetcher<
...
+            try {
+                const content = JSON.parse(object.content);
+                const contact: Partial<PseudoContact> = {
+                    hexpub: object.pubkey
+                };
+                if (content?.name) {
+                    contact.name = content.name;
+                } else if (content?.display_name) {
+                    contact.name = content.display_name;
+                }
```
benthecarman, line 360: in mutiny-node we do display_name then try name, should have the logic the same here

benthecarman (changes requested):

benthecarman:
this only searches by username, would be nice if it also worked with npubs

src/utils/fetchZaps.ts
futurepaul: yeah let's unify this somewhere but not now

benthecarman: In damus my `display_name` is Carman 🏴‍☠ but my `name` is `@benthecarman` so I definitely think we should use `display_name` here

futurepaul: find you convinced me

PR #817, opened 2024-01-12
prettier formatting consistency

nk1tz:
# What
- [x] ensure `format` script applies to all root files and `e2e/` as well as `src/`
- [x] include `prettier.config.mjs` in tsconfig to avoid eslint complaint in `prettier.config.mjs`
- [x] apply new format script
- [x] readme typo

Also  WHY `tabWidth: 4,` ? :D :D :D  my eyes...

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "check-types": "tsc --noemit",
         "eslint": "eslint src",
         "pre-commit": "pnpm run eslint && pnpm run check-types && pnpm run format",
-        "format": "prettier --write \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
+        "format": "prettier --write \"{.,src/**,e2e/**}/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
         "check-format": "prettier --check \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\""
     },
     "type": "module",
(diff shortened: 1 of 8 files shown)


benthecarman (approved):
lgtm, i think the iOS build failure is because you're not part of the org

package.json
```diff
@@ -9,7 +9,7 @@
         "check-types": "tsc --noemit",
         "eslint": "eslint src",
         "pre-commit": "pnpm run eslint && pnpm run check-types && pnpm run format",
-        "format": "prettier --write \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
+        "format": "prettier --write \"{.,src/**,e2e/**}/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
```
nk1tz, line 12: root + src + e2e
avoids linting node_modules etc..., can be added to easily
```
{.,src/**,e2e/**}
```

futurepaul:
> Also WHY `tabWidth: 4,` ? :D :D :D my eyes...

trying to discourage deeply nested anything... although I've recently heard a good argument for just using real tabs, so then the display of the tabs can be preferable per dev

futurepaul:
thank you for this!

since it seems somewhat related to what you're doing here, could you fix this warning for me?

Screenshot 2024-01-13 at 12 53 39 PM

futurepaul (approved):
thank you!

PR #818, opened 2024-01-13
Search Improvements

benthecarman:
Adds ability to send to users who only have a lnurl set.

Now shows users without any payment info, but disabled and greyed out.

Also switches to use a set over an array, this should make lookups faster but seems like most of the lag when typing is from the global search, maybe need to delay that by 250ms until the user is done typing

### DIFF
--- a/src/routes/Search.tsx
+++ b/src/routes/Search.tsx
@@ -91,10 +91,10 @@ function ActualSearch() {
         return (
             contacts()?.filter((c) => {
                 return (
-                    c.ln_address &&
-                    (c.name.toLowerCase().includes(s) ||
-                        c.ln_address?.toLowerCase().includes(s) ||
-                        c.npub?.includes(s))
+                    c.name.toLowerCase().includes(s) ||
+                    c.ln_address?.toLowerCase().includes(s) ||
+                    c.lnurl?.toLowerCase().includes(s) ||
+                    c.npub?.includes(s)
                 );
             }) || []
         );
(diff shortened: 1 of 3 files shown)


src/routes/Search.tsx
```diff
@@ -281,7 +281,15 @@ function ActualSearch() {
                         {(contact) => (
                             <button
                                 onClick={() => sendToContact(contact)}
-                                class="flex items-center gap-2"
+                                class={
```
futurepaul, line 284: if you're using logic for the style you want to use `classList`

put the default style in class, and then add

```
classList={{ "text-white/20 opacity-60": !contact.ln_address && !contact.lnurl }}
```

src/routes/Search.tsx
```diff
@@ -295,8 +303,19 @@ function ActualSearch() {
                                     <h2 class="overflow-hidden overflow-ellipsis text-base font-semibold">
                                         {contact.name}
                                     </h2>
-                                    <h3 class="overflow-hidden overflow-ellipsis text-sm font-normal text-neutral-500">
-                                        {contact.ln_address}
+                                    <h3
+                                        class={
```
futurepaul, line 307: classList here as well

src/routes/Search.tsx
```diff
@@ -444,7 +466,15 @@ function SingleContact(props: {
     return (
         <button
             onClick={() => createContactFromSearchResult(props.contact)}
-            class="flex items-center gap-2"
+            class={
```
futurepaul, line 469: this is a lot of copy pasting so we could probably make a `<ContactButton>` component file to dedupe

futurepaul (changes requested):

futurepaul (approved):

PR #819, opened 2024-01-13
Faster lookup of bip39 words

benthecarman:
Sets are faster than arrays for lookups: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#performance


futurepaul (approved):
nice, ty

PR #822, opened 2024-01-13
Fix editing amount on receive screen

### LINKED ISSUE #569: Enumerate all possible payment failure states

### LINKED ISSUE #681: Lightning receive problem when setting low amount
1. New wallet
2. Create an amount that is more than the minimum lightning receive
3. Create invoice
4. Go back one nav
5. Edit the amount to be less
6. Create invoice (should be on chain only)
7. Go back one nav
8. Edit the amount to be more than the minimum
9. Create invoice

Final result is that it says the amount is not high enough despite correcting it to be more than the previous minimum. Makes me think it's not using the new edited amount at all.

### LINKED ISSUE #820: Defaulting to On-chain message
When first creating an invoice for a small amount and getting the message "Defaulting to On-chain. Amount is too small for your initial Lightning receive." above the qr code, then going back one window, increasing the amount to for example 100k sats, creating a new invoice, the message "Defaulting to On-chain. Amount is too small for your initial Lightning receive." still appears even tough it's a big enough amount. Only if I go back to home screen and then creating a new invoice it switches to channel creation instead of on chain.

benthecarman:
Closes #820
Closes #681

Also fixes handling expired invoices

Should be the last thing needed to close #569

### DIFF
--- a/src/i18n/en/translations.ts
+++ b/src/i18n/en/translations.ts
@@ -137,6 +137,7 @@ export default {
         error_clipboard: "Clipboard not supported",
         error_keysend: "Keysend failed",
         error_LNURL: "LNURL Pay failed",
+        error_expired: "Invoice is expired",
         payment_pending: "Payment pending",
         payment_pending_description:
             "It's taking a while, but it's possible this payment may still go through. Please check 'Activity' for the current status.",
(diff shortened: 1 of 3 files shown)


futurepaul (approved):
wonderful thank you

PR #825, opened 2024-01-13
method chooser for send

### LINKED ISSUE #816: some way to choose send method with new design?
in #775 we lost the ability to choose send method (in the case of bip21 unified qr for instance). also we already don't have manual send method selection when a fedimint is added.

I'm thinking maybe a little icon, similar to the planned private / public button?

Screenshot 2024-01-11 at 4 54 05 PM

futurepaul:
closes #816 

https://github.com/MutinyWallet/mutiny-web/assets/543668/e91c25b2-da1e-4d5c-a1f8-a9b5ef90915e

big diff on `Swap.tsx` is due to deleting `<SafeArea>` so swap will fill the height of the screen nicer. actual change in swap is just changing `maxAmountSats` for the new style


src/routes/Send.tsx
```diff
@@ -637,6 +638,57 @@ export function Send() {
...
+
+        if (invoice()) {
+            return [lightningMethod()];
+        }
+
+        if (address()) {
+            return [onchainMethod()];
+        }
+
+        // We should never get here
```
benthecarman, line 672: should put an error log here so it is obvious we have a bug

benthecarman: this function doesn't handle keysends

src/components/AmountEditable.tsx
```diff
@@ -229,12 +234,53 @@ export const AmountEditable: ParentComponent<{
...
+        >
+            <AmountSats
+                amountSats={props.activeMethod.maxAmountSats!}
+                denominationSize="sm"
+                icon={
+                    props.activeMethod.method === "lightning"
+                        ? "lightning"
+                        : props.activeMethod.method === "onchain"
+                        ? "chain"
+                        : "community"
```
benthecarman, line 281: do we want this commutiny icon here?

benthecarman: might be better to have this just as a function next to the definition of `method` that coverts to an icon so we don't have to remember to change this too when we add another method

benthecarman (changes requested):

benthecarman (approved):

PR #832, opened 2024-01-16
Feedback clarification

TonyGiorgio:


benthecarman (approved):
Maybe should just remove the feedback form

TonyGiorgio:
> Maybe should just remove the feedback form

Honestly it's still valuable, just a small percentage of people that expect a follow up.

PR #834, opened 2024-01-17
Display pending nwc from dms

benthecarman:
needs https://github.com/MutinyWallet/mutiny-node/pull/967

Functionality is working, just needs proper design. I just made the sync nostr contacts button work with nsecs in the apps.

Will also detect nip-07 extensions


benthecarman:
In 844 now

PR #837, opened 2024-01-19
Lnurl withdrawal support

tompro:
Added very basic support for LNUrl withdrawal requests. I tried to just add the business logic and leave all graphical stuff untouched when possible (you probably don't want me to mess up your CI). 

The following UI related things will need some follow up:
- Add a paste field in the receive screen (LNUrl withdrawal can only be started from scan/paste screen as of now)
- Show a more telling loading screen while withdrawal is running. Right now it just shows a QR with checking (LNUrl can take a while)

Another thing I could not solve is to preset an fixed amount for the withdrawal (when min === max). Whenever I set the initial amount in the amount editor it is immediately reset to 0 from the editor (invoking the callback). I did not mess with that central component.

### DIFF
--- a/src/i18n/en/translations.ts
+++ b/src/i18n/en/translations.ts
@@ -113,7 +113,8 @@ export default {
             gift: "Lightning Gift"
         },
         remember_choice: "Remember my choice next time",
-        what_for: "What's this for?"
+        what_for: "What's this for?",
+        lnurl_withdrawal_in_progress: "LNUrl withdrawal in progress"
     },
     send: {
         search: {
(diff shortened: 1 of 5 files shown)


tompro:
Think I found a way to solve the fixed amount issue mentioned above

benthecarman:
> what's a good way for me to test this with mutinynet?

there's one on the faucet

```
lnurl1dp68gurn8ghj7enpw43k2apwd46hg6tw09hx2apwvdhk6tmpwp5j7mrww4excac7utxd6
```

tompro:
> what's a good way for me to test this with mutinynet?

I can provide you with some fixed amt ones (skips enter amt) from our dev server. Needs to be DM though as they are one time use. Just let me know.

futurepaul:
closing in favor of #905. thank you for doing this! it made really straightforward for me to do in my own style, made you a co-author on #905

PR #838, opened 2024-01-19
Display individual fedimint balances

benthecarman:
Not sure if the switch / match is the cleanest way to do this but it works :man_shrugging: 

image


futurepaul (approved):
works great, there are probably other ways to write this but this is clean enough to merge imo

PR #842, opened 2024-01-22
feat: add language selection on settings page

### LINKED ISSUE #830: feature: add language options on settings page
## What
I've discussed this with @futurepaul and @benthecarman in the past few days, and I think it would be useful to be able to change the language through the settings page as well, instead of just relying on the browser/phone default language.

## Why
It's useful for users who would like to have different languages on their devices for different purposes or are abroad and another language would be suitable in that situation.

## How
Something alongside the style and options for currency selection, nothing else would be needed to change in my opinion.

--

**note:** I'm open to trying this one but would like to know your views and opinions beforehand.

oleonardolima:
fixes #830 
closes #875

### DIFF
--- a/src/components/ChooseLanguage.tsx
+++ b/src/components/ChooseLanguage.tsx
@@ -1,0 +1,95 @@
+import { createForm } from "@modular-forms/solid";
+import { useNavigate } from "@solidjs/router";
+import { createSignal, For, Show } from "solid-js";
+
+import { Button, ExternalLink, InfoBox, NiceP, VStack } from "~/components";
+import { useI18n } from "~/i18n/context";
+import { useMegaStore } from "~/state/megaStore";
+import { eify, EN_OPTION, Language, LANGUAGE_OPTIONS, timeout } from "~/utils";
+
+type ChooseLanguageForm = {
+    selectedLanguage: string;
+};
+
+const COMBINED_OPTIONS: Language[] = [EN_OPTION, ...LANGUAGE_OPTIONS];
+
+export function ChooseLanguage() {
+    const i18n = useI18n();
+    const [error, setError] = createSignal<Error>();
+    const [state, actions] = useMegaStore();
+    const [loading, setLoading] = createSignal(false);
+    const navigate = useNavigate();
+
+    const [_chooseLanguageForm, { Form, Field }] =
+        createForm<ChooseLanguageForm>({
+            initialValues: {
+                selectedLanguage: state.lang ?? ""
+            },
+            validate: (values) => {
+                const errors: Record<string, string> = {};
+                if (values.selectedLanguage === undefined) {
+                    errors.selectedLanguage = i18n.t(
+                        "settings.language.error_unsupported_language"
+                    );
+                }
+                return errors;
+            }
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        try {
+            actions.saveLanguage(f.selectedLanguage);
+
+            await i18n.changeLanguage(f.selectedLanguage);
+
+            await timeout(1000);
+            navigate("/");
+        } catch (e) {
+            console.error(e);
+            setError(eify(e));
+            setLoading(false);
+        }
+    };
+
+    return (
+        <VStack>
+            <Form onSubmit={handleFormSubmit} class="flex flex-col gap-4">
+                <NiceP>{i18n.t("settings.language.caption")}</NiceP>
+                <ExternalLink href="https://github.com/MutinyWallet/mutiny-web/issues/new">
+                    {i18n.t("settings.language.request_language_support_link")}
+                </ExternalLink>
+                <div />
+                <VStack>
+                    <Field name="selectedLanguage">
+                        {(field, props) => (
+                            <select
+                                {...props}
+                                value={field.value}
+                                class="w-full rounded-lg bg-m-grey-750 py-2 pl-4 pr-12 text-base font-normal text-white"
+                            >
+                                <For each={COMBINED_OPTIONS}>
+                                    {({ value, shortName }) => (
+                                        <option
+                                            selected={field.value === shortName}
+                                            value={shortName}
+                                        >
+                                            {value}
+                                        </option>
+                                    )}
+                                </For>
+                            </select>
+                        )}
+                    </Field>
+                    <Show when={error()}>
+                        <InfoBox accent="red">{error()?.message}</InfoBox>
+                    </Show>
+                    <div />
+                    <Button intent="blue" loading={loading()}>
+                        {i18n.t("settings.language.select_language")}
+                    </Button>
+                </VStack>
+            </Form>
+        </VStack>
+    );
+}
--- a/src/state/megaStore.tsx
+++ b/src/state/megaStore.tsx
@@ -54,6 +54,7 @@ type MegaStore = [
         price_sync_backoff_multiple?: number;
         price: number;
         fiat: Currency;
+        lang?: string;
         has_backed_up: boolean;
         wallet_loading: boolean;
         setup_error?: Error;
@@ -83,6 +84,7 @@ type MegaStore = [
         checkForSubscription(justPaid?: boolean): Promise<void>;
         fetchPrice(fiat: Currency): Promise<number>;
         saveFiat(fiat: Currency): void;
+        saveLanguage(lang: string): void;
         saveNpub(npub: string): void;
         setPreferredInvoiceType(
             type: "unified" | "lightning" | "onchain"
@@ -132,6 +134,7 @@ export const Provider: ParentComponent = (props) => {
         load_stage: "fresh" as LoadStage,
         settings: undefined as MutinyWalletSettingStrings | undefined,
         safe_mode: searchParams.safe_mode === "true",
+        lang: localStorage.getItem("i18nexLng") || undefined,
         npub: localStorage.getItem("npub") || undefined,
         preferredInvoiceType: "unified" as "unified" | "lightning" | "onchain",
         betaWarned: localStorage.getItem("betaWarned") === "true",
@@ -331,6 +334,10 @@ export const Provider: ParentComponent = (props) => {
                 price: price,
                 fiat: fiat
             });
+        },
+        saveLanguage(lang: string) {
+            localStorage.setItem("i18nextLng", lang);
+            setState({ lang });
         },
         saveNpub(npub: string) {
             localStorage.setItem("npub", npub);
--- a/src/utils/languages.ts
+++ b/src/utils/languages.ts
@@ -1,0 +1,20 @@
+export interface Language {
+    value: string;
+    shortName: string;
+}
+
+export const EN_OPTION: Language = {
+    value: "English",
+    shortName: "en"
+};
+
+export const LANGUAGE_OPTIONS: Language[] = [
+    {
+        value: "Português",
+        shortName: "pt"
+    },
+    {
+        value: "Korean",
+        shortName: "ko"
+    }
+];
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -9,6 +9,7 @@ export * from "./AmountEditable";
 export * from "./BalanceBox";
 export * from "./BetaWarningModal";
 export * from "./ChooseCurrency";
+export * from "./ChooseLanguage";
 export * from "./ContactEditor";
 export * from "./ContactForm";
 export * from "./ContactViewer";
--- a/src/i18n/config.ts
+++ b/src/i18n/config.ts
@@ -32,8 +32,10 @@ const i18n = use(LanguageDetector).init(
         fallbackNS: false,
         debug: true,
         detection: {
-            order: ["querystring", "navigator", "htmlTag"],
-            lookupQuerystring: "lang"
+            order: ["localStorage", "querystring", "navigator", "htmlTag"],
+            lookupQuerystring: "lang",
+            lookupLocalStorage: "i18nextLng",
+            caches: ["localStorage"]
         },
         resources: resources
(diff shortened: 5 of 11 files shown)


oleonardolima:
@futurepaul I'm curious about how I should override the i18n context, is it already being done on another part of the code?

oleonardolima:
> @futurepaul I'm curious about how I should override the i18n context, is it already being done on another part of the code?

As we discussed, it would need a new language field and function update, right ?

Something like this:

https://github.com/MutinyWallet/mutiny-web/blob/c680ab2805ac72399b38e1ed88fbe3d56a1f1544/src/state/megaStore.tsx#L327-L334

futurepaul:
> Something like this:

yeah that looks right to me. you set it in the state, and then you reference that value from the global state to pass to the config during i18npovider (https://github.com/MutinyWallet/mutiny-web/blob/master/src/components/I18nProvider.tsx)

it looks like i18n has its cachuserlanguage thing but probably don't need to bother with it would rather just have it stored in megastore https://github.com/i18next/i18next-browser-languageDetector

oleonardolima:
> > Something like this:
> 
> yeah that looks right to me. you set it in the state, and then you reference that value from the global state to pass to the config during i18npovider (https://github.com/MutinyWallet/mutiny-web/blob/master/src/components/I18nProvider.tsx)
> 
> it looks like i18n has its cachuserlanguage thing but probably don't need to bother with it would rather just have it stored in megastore https://github.com/i18next/i18next-browser-languageDetector

nice! thx

oleonardolima:
> ....
> 
> it looks like i18n has its cachuserlanguage thing but probably don't need to bother with it would rather just have it stored in megastore https://github.com/i18next/i18next-browser-languageDetector

yep, I've tried using the cache and detection options, but doesn't to work 🤔, I'll try getting it directly from megaStore then lol

benalleng:
I see the issue here! you need to modify the detection in the i18n config as such
```
        detection: {
            order: ["querystring", "localStorage", "navigator", "htmlTag"],
            lookupQuerystring: "lang",
            lookupLocalStorage: "i18nextLng",
            caches: ["localStorage"],
        },
```
This will allow the localStorage item to be found

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    timeout,
+    LANGUAGE_OPTIONS,
+    EN_OPTION
+} from "~/utils";
+
+type ChooseLanguageForm = {
+    selectedLanguage: string;
+};
+
+const COMBINED_OPTIONS: Language[] = [...LANGUAGE_OPTIONS];
```
benalleng, line 20: I think we should add english as an option so maybe just `[EN_OPTION, ...LANGUAGE_OPTIONS];`

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,101 @@
...
+            }
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        try {
+            actions.saveLanguage(findLanguageByValue(f.selectedLanguage).label);
+
+            await timeout(1000);
+            navigate("/");
```
benalleng, line 58: One issue I can't figure out is how to refresh the i18n object, since `navigate("/")` doesn't force a refresh I think a `window.location.reload()` could be an option unless there is some i18n api call I am missing to force a locale refresh

oleonardolima:
> I see the issue here! you need to modify the detection in the i18n config as such
> 
> ```
>         detection: {
>             order: ["querystring", "localStorage", "navigator", "htmlTag"],
>             lookupQuerystring: "lang",
>             lookupLocalStorage: "i18nextLng",
>             caches: ["localStorage"], // not sure if this specific key is necessry
>         },
> ```
> 
> This will allow the localStorage item to be found

Nice! I've tried that, but the problem remained I think mainly because it wasn't being refreshed 🤔

src/components/ChooseLanguage.tsx
oleonardolima: Cool, that's better.

src/components/ChooseLanguage.tsx
oleonardolima: Yep, I didn't find any specific i18n call, but I'll try with the windown.location.reload() then 🤔

benalleng: Ideally we don't want to have the page refresh, perhaps there is a way to make the i18nProvider component re-render on a language change

oleonardolima: Cool, I'll do a little bit more research about it then.

oleonardolima: @benthecarman gave me a little help and we figured out 🚀

benalleng: awesome works great!

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,103 @@
...
+    timeout,
+    LANGUAGE_OPTIONS,
+    EN_OPTION
+} from "~/utils";
+
+type ChooseLanguageForm = {
+    selectedLanguage: string;
+};
+
+const COMBINED_OPTIONS: Language[] = [...LANGUAGE_OPTIONS];
```
benalleng, line 20: I still think we should add in the `EN_OPTION` here so that a user can switch back to english

benalleng (changes requested):
Really close! great work, just some small things and then make sure to run `pnpm pre-commit` to clean up any prettier formatting

src/state/megaStore.tsx
```diff
@@ -30,6 +30,7 @@ import {
     BTC_OPTION,
     Currency,
     eify,
+    Language,
```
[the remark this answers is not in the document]
oleonardolima: Oh, done

src/components/ChooseLanguage.tsx
oleonardolima: Sure, I forgot to add it.

oleonardolima:
> Really close! great work, just some small things and then make sure to run `pnpm pre-commit` to clean up any prettier formatting

@benalleng I was talking with @benthecarman about adding some non-blocking step on CI for showing that some translations are missing, wdyt ?

benalleng (approved):

benalleng:
Definitely! I think we should probably make a new PR for that CI check though

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,99 @@
...
+                <VStack>
+                    <Field name="selectedLanguage">
+                        {(field, props) => (
+                            <select
+                                {...props}
+                                value={field.value}
+                                class="w-full rounded-lg bg-m-grey-750 py-2 pl-4 pr-12 text-base font-normal text-white"
+                            >
+                                <For each={COMBINED_OPTIONS}>
+                                    {({ value, label }) => (
```
benalleng, line 81: making this `_label` should fix this

oleonardolima:
> Definitely! I think we should probably make a new PR for that CI check though

I'll add some playwright tests on this PR as well, and then we are good to go :)

src/utils/languages.ts
```diff
@@ -0,0 +1,24 @@
+export interface Language {
+    value: string;
+    label: string;
+    hasSymbol?: string;
```
benalleng, line 4: That unused arg warning also brought t my attention that we don't seem to really need `hasSymbol`. unlike currency I don't think the symbol is super import for identifying the current language like it is for currency

src/components/ChooseLanguage.tsx
oleonardolima: Good, I'll do it.

src/utils/languages.ts
oleonardolima: Good catch! I left it there since the beginning, removing it.

src/utils/languages.ts
benalleng, line 1: from the element chat we came to some consensus that we might want to change the key names for this interface as it might be a little confusing with the `value` being the label and the `label` being the value sent to i18next.  @futurepaul suggested we use `shortName` to replace label

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,99 @@
...
+            }
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        try {
+            actions.saveLanguage(findLanguageByValue(f.selectedLanguage).label);
+
+            await i18n.changeLanguage(
+                findLanguageByValue(f.selectedLanguage).label
```
benalleng, line 52: if we make the change the would also need to be `shortName` instead of `label`

src/state/megaStore.tsx
```diff
@@ -132,6 +134,7 @@ export const Provider: ParentComponent = (props) => {
         load_stage: "fresh" as LoadStage,
         settings: undefined as MutinyWalletSettingStrings | undefined,
         safe_mode: searchParams.safe_mode === "true",
+        lang: localStorage.getItem("lang") || undefined,
```
benalleng, line 137: this should be 
```
lang: localStorage.getItem("i18nexLng") || undefined,
```

benalleng (commented):
Made another pass after I had a chance to look at #875 and the `value` <=> `label` stuff was most of the problem with the select options being weird but I did catch one small change in addition to that

src/components/ChooseLanguage.tsx
benalleng: I think it would be easier to look at #875 to make sense of this

oleonardolima: Cool, I'll take a look at it

src/state/megaStore.tsx
oleonardolima: Nice, thanks!

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,104 @@
...
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        try {
+            actions.saveLanguage(findLanguageByValue(f.selectedLanguage).value);
+
+            await i18n.changeLanguage(
+                findLanguageByValue(f.selectedLanguage).shortName
+            );
```
benthecarman, line 53: nit: we're calling `findLanguageByValue` twice here, could do `const lang = findLanguageByValue(f.selectedLanguage)` then use the result for the two calls below

benthecarman (approved):

src/components/ChooseLanguage.tsx
benalleng: if we change the <option> params like so we actually don't need the `findLanguageByValue()` function at all
```
<option
        selected={field.value === shortName}
        value={shortName}
    >
```

oleonardolima: Good, done!

benthecarman (approved):

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,103 @@
...
+                                {...props}
+                                value={field.value}
+                                class="w-full rounded-lg bg-m-grey-750 py-2 pl-4 pr-12 text-base font-normal text-white"
+                            >
+                                <For each={COMBINED_OPTIONS}>
+                                    {({ value, shortName }) => (
+                                        <option
+                                            selected={field.value === shortName}
+                                            value={value}
+                                        >
```
benalleng, line 84: The value changed needs to be the shortName 
```
value={shortName}
```

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        const lang = findLanguageByValue(f.selectedLanguage);
+        try {
+            actions.saveLanguage(lang.value);
+
+            await i18n.changeLanguage(lang.shortName);
```
benalleng, line 52: the lang state and i18n api call should be the same key, that is why in the megaStore we are initializing with `localStorage.getItem('i18nextLng')`. since we are using the `shortName as the value we can just point to the selectedLanguage
```
actions.saveLanguage(f.selectedLanguage);

await i18n.changeLanguage(f.selectedLanguage);
```

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,103 @@
...
+        return (
+            COMBINED_OPTIONS.find((language) => language.value === value) ??
+            EN_OPTION
+        );
+    }
+
+    const [_chooseLanguageForm, { Form, Field }] =
+        createForm<ChooseLanguageForm>({
+            initialValues: {
+                selectedLanguage: findLanguageByValue(state.lang ?? "").value
```
benalleng, line 33: Same here we can simply point to the current state since the megaStore initializes with the 'i18nextLng' key
```
selectedLanguage: state.lang ?? ""
```

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const [state, actions] = useMegaStore();
+    const [loading, setLoading] = createSignal(false);
+    const navigate = useNavigate();
+
+    function findLanguageByValue(value: string) {
+        return (
+            COMBINED_OPTIONS.find((language) => language.value === value) ??
+            EN_OPTION
+        );
+    }
```
benalleng, line 28: since we are just using the shortName everywhere and the value is used as a label in the UI only this can be removed

benalleng (changes requested):
Sorry to butt heads on this! But switching back and forth with the value and findLanguageByValue() is still causing problems.

If you start in another language it still always displays english as the default

benalleng (approved):
tACK great work on this!

PR #844, opened 2024-01-22
social redesign

futurepaul:


futurepaul:
rebased to master, relies on https://github.com/MutinyWallet/mutiny-node/pull/992

TonyGiorgio:
Looks like it needs a rebase and maybe an pnpm lock update?

TonyGiorgio (changes requested):
This feels awesome, really nice work. I made some notes testing it: 

- Should the backup prompt be above the contacts search? Looks a little broken up currently.
- For the empty hints, should clicking on the hints themselves do the action? 
- Quick actions icon don't work on either the home page or in messages
- Make message sending instant, instead of blocking. Just have a pending icon instead of saying 'Just now' until it sends.
- For "everyone" should we pull from the global feed? Or change that to "friends" instead of everyone since that's just the local friend list? Maybe a call to action (eventually) to follow some people instead of a blank list.
- Tags are not vertically centered in their pill bubble
- Remove the lightning address "elegant-giraffe@mutiny.plus" part for now, replace with an "add lightning address" link that does nothing for now?
- For the nostr npub at the bottom, maybe say "Nostr Identity" instead? Just to give some indication about what it is. 
- In DMs, "send" should be grayed out if they do not have a lightning address
- In DMs, paying a normal invoice should not allow the option of anon vs private switcher, it should just be private since it's just an invoice
- In a follow up, would it be possible to show a notification for friends that have recently messaged you? I think that'll be important without having to click into the DM first

futurepaul:
addressed most of tony's comments above.

except for:

- [ ] In a follow up, would it be possible to show a notification for friends that have recently messaged you? I think that'll be important without having to click into the DM first (this would be fun)
- [ ] Make message sending instant, instead of blocking. (this is hard to do without causing a flicker that feels worse, at least in my attempts so far)

noticed a few things along the way:
- [ ] somehow broke loading my chat with mutinynet. loading other chats works. I think it's a mutiny-node problem no 100% sure though (@benthecarman)
- [ ] can end up on a send screen and clicking "back" doesn't work??
- [ ] for getting the zap feed I should ask the backend "what's my contact list?" (instead of asking primal api)
- [ ] minifab menu on desktop isn't above the minifab button

benthecarman:
- On firefox I am just getting a blank screen, tried clearing storage and its still happens. (its not my dark mode plugin).
- On a fresh wallet the first time I hit receive it took awhile for the USD amount to show up
- Something we talked about before, not sure if we want to get in here, but removing the password after backup could be a little nice
- instinctively I click on people's faces in my activity when trying to send to them, maybe the face should go to chat, or on the activity item view, should have a link to get to the chat
- definitely agree on pressing enter to send in dms
- should we have some of CTA inside the empty chat state, it feels weird with nothing there
- see a couple weird zaps in the friends feed like this, if i click on their face it doesn't take to the chat like the others do image

benthecarman:
Not sure what could be going wrong with paul's dms, maybe he's sent to many something is crashing?

TonyGiorgio:
> instinctively I click on people's faces in my activity when trying to send to them, maybe the face should go to chat

I did the same

benthecarman:
> Might be private zaps or directly from a minibits wallet or something

that was my assumption, there were some showing up as private in my feed though, we probably just aren't handling some properly

futurepaul:
> * On firefox...

sweet thank you for this list, will work through these today. I've been patching up the tests because a lot of nav has changed that breaks the tests

futurepaul:
okay got most of these @benthecarman (with your help)

- [x]  instinctively I click on people's faces in my activity when trying to send to them, maybe the face should go to chat, or the activity item view should have a link to get to the chat, maybe both
- [x]  definitely agree on pressing enter to send in dms
- [x] should we have some of CTA inside the empty chat state, it feels weird with nothing there
- [x]  see a couple weird zaps in the friends feed like this, if i click on their face it doesn't take to the chat like the others do

I know you're still looking into
- [ ] anon zaps not showing up

futurepaul:
I gotta update all the tests for the new ui. so far I have

- [x] routes
- [ ] encrypt
- [ ] fedimint
- [ ] load
- [ ] restore
- [ ] roundtrip

futurepaul:
Screenshot 2024-02-22 at 4 18 17 PM

futurepaul:
found a bunch of little ui problems when testing on mobile:

- [ ] overflow scroll (my old nemesis)
- [ ] balance isn't centered
- [ ] the "shiny" edges of buttons are thicker on mobile? (maybe need 0.5px on sm screens?)
- [ ] don't suggest autofill in search (old problem)
- [ ] can I get rid of the weird camera view resize flicker? (old problem)
- [ ] can I make the multiple tab screen less ugly? (always see it when moving seed to new device)
- [ ] it's time to toggle hide balance we've waited long enough
- [ ] really bugs me that the amount editor screens don't squeeze down when the keyboard is open
- [ ] feels a little empty with just one tx and all the prompts gone?

IMG_1912

- [ ] need to add nsec import if I'm not gonna have setup screen

TonyGiorgio:
- There's a few places where it says "your activity is private by default" when it comes to nostr profile information. This is not true and should be removed. 

Screenshot 2024-03-13 at 3 55 47 PM

- The text at the top in general is redundant and unnecessary. 

- It should probably let people set a lightning address in general until it's ready. Also if they are not subscribers then they don't get a mutiny one but can set the lightning address from another place? 

- The federation QR code is showing up as undefined. This is not a problem on signet currently. @benthecarman is this a hotfix problem? The federation invite code didn't get applied to v0.6.0-rc? 

- should both of these be here for the same payment?
Screenshot 2024-03-13 at 4 02 00 PM

- there's not a flow for adding friends. This takes the user to a search screen, where they can "find" a contact but can't do anything with them except pay or request funds. Clicking on one of these users will 'add to contact' but doesn't actually add them in any meaningful way. I think there needs to be an "add friend" button somewhere when searching or finding a contact. Remember, we're a nostr client now, and one where this might be their first client. Adding friends/following is a necessary part of that social functionality.
Screenshot 2024-03-13 at 4 04 14 PM

- on that note, I say we remove the npub import completely and force npriv import or nothing at all. If we want to remove browser extension support, that's fine too. Paul has been struggling with this and if it helps in any way I say we remove anything that isn't going to be covered for a majority of new users with the focus on the ios/android app.

- there's something about these home screen call to action "buttons" that they don't really feel like buttons. And when they are stacked up, it is not very visually appealing. This will be what everyone lands on, can we make it look and feel better? The "secure funds" one should be way more pressing for urgency at the very least. It just feels like it blends in.

Screenshot 2024-03-13 at 4 11 01 PM

---

App feels pretty good, nice work! Mostly visual/cleanup things but one core functionality that I think should be in this release is adding friends.

TonyGiorgio:
> > Yeah we haven't cut an rc in awhile
> 
> Having some trouble with the fedimint test on the send timing out, curious if a rc3 would fix that

can you run it on master and see? What specifically is the error?

futurepaul:
> can you run it on master and see? What specifically is the error?

ran it on master, similar flakiness so doesn't seem to be a mutiny-node problem

benthecarman:
Something I noticied when testing. This is an on-chain payment but it says I have 0 balance and the activity item has a lightning bolt.

Maybe the balance at the top should be an aggregate or at least display the highest.

image

benthecarman:
When I click on an activity item the UI in the background gets squished into the corner

image

TonyGiorgio (approved):
Huge, very well done on this. Feel free to make any small patches that you like and then merge.

benthecarman (approved):
:ship: it

PR #850, opened 2024-01-25
Prod v0.5.7

TonyGiorgio:
Needs another round of full end to end testing first but getting the PR ready.


benthecarman (approved):
old nwc's seem to not be working well but seems like a blastr error, this update gets us off of it anyways :man_shrugging:

PR #852, opened 2024-01-26
Warn about hodl invoice

benthecarman:
Before we only warned for zeuspay lightning addresses, now we warn for the actual invoices as well

### DIFF
--- a/src/routes/Send.tsx
+++ b/src/routes/Send.tsx
@@ -1,4 +1,4 @@
-import { MutinyInvoice } from "@mutinywallet/mutiny-wasm";
+import { MutinyInvoice, MutinyWallet } from "@mutinywallet/mutiny-wasm";
 import { A, useNavigate, useSearchParams } from "@solidjs/router";
 import {
     createEffect,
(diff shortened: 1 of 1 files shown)


src/routes/Send.tsx
```diff
@@ -451,11 +451,15 @@ export function Send() {
...
         state.mutiny_wallet
             ?.decode_invoice(source.invoice!)
-            .then((invoice) => {
+            .then(async (invoice) => {
                 if (invoice.expire <= Date.now() / 1000) {
                     navigate("/search");
                     throw new Error(i18n.t("send.error_expired"));
                 }
+                const isHodl = await MutinyWallet.is_potential_hodl_invoice(
+                    source.invoice!
```
TonyGiorgio, line 460: why is source here and not not `invoice`?

src/routes/Send.tsx
```diff
@@ -451,11 +451,15 @@ export function Send() {
     function processInvoice(source: ParsedParams & { invoice: string }) {
         state.mutiny_wallet
             ?.decode_invoice(source.invoice!)
-            .then((invoice) => {
+            .then(async (invoice) => {
                 if (invoice.expire <= Date.now() / 1000) {
                     navigate("/search");
                     throw new Error(i18n.t("send.error_expired"));
                 }
+                const isHodl = await MutinyWallet.is_potential_hodl_invoice(
```
TonyGiorgio, line 459: I think this should just happen below in a single spot where it is being set. Now you have it here twice. I think this method can just replace the previous.

TonyGiorgio (changes requested):

src/routes/Send.tsx
benthecarman: invoice isn't a string, it is a `MutinyInvoice`

benthecarman:
Not needed

PR #858, opened 2024-01-29
use primal cdn for images

futurepaul:
this uses the `s` (small) sized images from primal for contacts which should speed up loads. also because it's all from one origin I think the browser can batch these better.

question: is there a way to do the same with our own cache url?

### DIFF
--- a/src/routes/Search.tsx
+++ b/src/routes/Search.tsx
@@ -465,6 +465,16 @@ function ContactButton(props: {
     onClick: () => void;
 }) {
     const i18n = useI18n();
+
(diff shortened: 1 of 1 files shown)


benthecarman (commented):
We use the contact image url in a bunch of places, should be done there too. Maybe it'd be easier if we actually did this on the backend?

futurepaul:
search was the main place it was really hurting me, but yeah if you want to rewrite this in the background (or add a separate field?) that might be good. as long as it's not going to complicate sync too much

benthecarman (approved):

PR #862, opened 2024-02-01
Fedimint sweep

TonyGiorgio:
This does a few refactors on common components. 

Does the swap and warns the user when a channel fee may be required. However one critical part is to use `estimate_sweep_federation_fee` once the user has selected the amount so that they can see the fee before they continue. Handles a bit different than the lightning->channel open flow does today which is hard for me to replicate. 

Relies on https://github.com/MutinyWallet/mutiny-node/pull/995

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "com.mutinywallet.mutinywallet"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 48
-        versionName "0.5.7"
+        versionCode 49
+        versionName "0.5.8"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
              // Files and dirs to omit from the packaged assets dir, modified to accommodate modern web apps.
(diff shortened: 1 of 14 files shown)


TonyGiorgio:
Known problems: 
- putting in a small amount and trying to do fee estimation when a new channel is needed doesn't work properly (amount is too low)
- i don't yet handle when the user previews the fee and then wants to change the fee, it should ask for another preview and don't allow edits (or switch back to preview button instead of confirm button)
- I think when a new channel is needed, the final success screen shows the amount that was transfered on the fedimint side (minus fedimint fee) but that amount is not considering the channel open fee. However the fee is accurate I think. Needs to investigate more, but perhaps that's a bug on the mutiny node side.

futurepaul:
@TonyGiorgio okay I think I've cleaned this up / addressed your issues, let me know if I'm missing anything

TonyGiorgio:
Awesome thank you so much. I'll test tonight.

TonyGiorgio:
Only frontend issue here is not being able to hit the back button on desktop to edit the amount

benthecarman:
2 issues I noticed

1. No max button, @TonyGiorgio said that @futurepaul removed this from the codebase, imo we should bring it back, if that is true though can be a follow up
2. After a successful swap it says `Swap Initiated` and that the balance will be added, however, it should be added already since the invoice was paid
image

TonyGiorgio:
> No max button, @TonyGiorgio said that @futurepaul removed this from the codebase, imo we should bring it back, if that is true though can be a follow up
I do think this should be a follow up. 

> After a successful swap it says Swap Initiated and that the balance will be added, however, it should be added already since the invoice was paid

This has been fixed

futurepaul (approved):
tony the frontend master

PR #867, opened 2024-02-06
Refetch federation balances on updates

### LINKED ISSUE #853: Refresh federation balance when federations are added or removed
If you add a federation, put funds in it, and then remove it, after https://github.com/MutinyWallet/mutiny-node/pull/995 it should have a balance if you add it again without refreshing. It is uses the balances it computed at the beginning of the page load. It should refresh whenever a federation is added.

benthecarman:
Closes #853


futurepaul (approved):
worked for me

TonyGiorgio:
What does this do and what's the problem?

PR #870, opened 2024-02-06
Prod v0.5.8

TonyGiorgio:


futurepaul:
we can get https://github.com/MutinyWallet/mutiny-web/pull/867 in too?

TonyGiorgio:
> we can get https://github.com/MutinyWallet/mutiny-web/pull/867 in too?

No, I'm only now seeing it. Should have notified me before all the testing.

benthecarman (approved):
imo #867 isn't really a big deal, only matters while adding a federation

futurepaul (approved):

PR #875, opened 2024-02-10
Try to fix lang select

futurepaul:


src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const [state, actions] = useMegaStore();
+    const [loading, setLoading] = createSignal(false);
+    const navigate = useNavigate();
+
+    function findLanguageByValue(value: string) {
+        return (
+            COMBINED_OPTIONS.find((language) => language.value === value) ??
+            EN_OPTION
+        );
+    }
```
benalleng, line 28: This function is trying to find the `Language` object based on the value that is stored in localStorage or state which is acutally the `shortName` not the value displayed to the user in this component
```
function findLanguageByShortName(value: string) {
    return (
        COMBINED_OPTIONS.find((language) => language.shortName === shortName) ??
        EN_OPTION
    );
}
```

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const [_chooseLanguageForm, { Form, Field }] =
+        createForm<ChooseLanguageForm>({
+            initialValues: {
+                selectedLanguage: findLanguageByValue(state.lang ?? "").value
```
benalleng, line 33: this is selecting the `value` as the default, so this is the only place where we are actually using `value` everywhere else is `shortName`
```
selectedLanguage: findLanguageByShortName(state.lang ?? "").value
```

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,104 @@
...
+            }
+        });
+
+    const handleFormSubmit = async (f: ChooseLanguageForm) => {
+        setLoading(true);
+        try {
+            actions.saveLanguage(findLanguageByValue(f.selectedLanguage).value);
+
+            await i18n.changeLanguage(
+                findLanguageByValue(f.selectedLanguage).label
```
benalleng, line 52: these values are saving the state and passing the new language to the i18n api and setting the new state so again they should be using the `shortName`
```
actions.saveLanguage(findLanguageByShortName(f.selectedLanguage).shortName);

await i18n.changeLanguage(
    findLanguageByShortName(f.selectedLanguage).shortName
);
```

benalleng (changes requested):
The `value` `label` confusion was causing problems everywhere it seems! If we switch over to the `shortName` key I its a lot easier to make sense of these changes.

src/components/ChooseLanguage.tsx
```diff
@@ -0,0 +1,104 @@
...
+                            <select
+                                {...props}
+                                value={field.value}
+                                class="w-full rounded-lg bg-m-grey-750 py-2 pl-4 pr-12 text-base font-normal text-white"
+                            >
+                                <For each={COMBINED_OPTIONS}>
+                                    {({ value }) => (
+                                        <option
+                                            selected={field.value === value}
+                                            value={value}
```
benalleng, line 84: the value we are looking to change is actually the shortName and not the value, so
```
value={shortName}
```

PR #881, opened 2024-02-13
Update translation keys

benalleng:
Began working with someone to update the korean translation and figured I should go through and try to catch some stray text that was not using translation keys

also adds a small fallback enhancement for the language selector so that if there is no language state it checks if there is a i18n object instead of an empty string
```
selectedLanguage: state.lang ?? i18n.language
```


futurepaul (approved):
thank you!

PR #883, opened 2024-02-13
Bump to 0.5.9

benthecarman:


TonyGiorgio (changes requested):
No go, BDK sync regression.

TonyGiorgio (approved):
Doesn't seem like a regression on the BDK side and more from the fact that I have had many channels and it was doing a ton of syncs anyways. There's open issues on cutting that down on the LDK level.

PR #888, opened 2024-02-16
Remove beta

futurepaul:
these are fixes requested by apple review


benthecarman (changes requested):
How about we just turn these off if we're in ios

futurepaul:
I'd rather just delete. it's either beta or it's not

benthecarman (approved):
Could make the argument pwa is more risky than native app, but I guess it's fine with me

PR #892, opened 2024-02-19
CI warnings for keys diff from fallback lang

benalleng:
The way I went about this adds a whole new folder for internal eslint rules as I couldn't figure out how to add a single rule through a vite project. Additionally I could be far overcomplicating it with an eslint rule all together but I only found pre-made eslint plugins for json files not typescript files. At the end of the day the logic works but I would still like to try and simplify it.

ultimately this rule will be disabled exept when a translator wants to check their work they can switch it in the .eslintrc.cjs file from "off" => "warn" or whatever

while it doesn't use line numbers for the warning it specifies the key path for the dev to track down the missing or extra key
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
         "build": "tsc && vite build",
         "check-types": "tsc --noemit",
         "eslint": "eslint src",
-        "pre-commit": "pnpm run eslint && pnpm run check-types && pnpm run format",
+        "eslint-path": "eslint",
+        "check-i18n-keys": "eslint src/i18n/**/translations.ts --rule \"{internal-rules/check-i18n-keys: warn}\"",
+        "pre-commit": "pnpm eslint && pnpm run check-types && pnpm run format",
         "format": "prettier --write \"{.,src/**,e2e/**}/*.{ts,tsx,js,jsx,json,css,scss,md}\"",
         "check-format": "prettier --check \"src/**/*.{ts,tsx,js,jsx,json,css,scss,md}\""
     },
@@ -26,6 +28,7 @@
         "eslint": "^8.52.0",
         "eslint-import-resolver-typescript": "2.7.1",
         "eslint-plugin-import": "2.27.5",
+        "eslint-plugin-internal-rules": "file:tools/internal-rules",
         "eslint-plugin-prettier": "4.2.1",
         "eslint-plugin-solid": "0.13.0",
         "postcss": "^8.4.31",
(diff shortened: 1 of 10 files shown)


benalleng:
Yeah i noticed it seems to be only running on like the first 10 lines of korean on the github diff but locally it works everywhere

benalleng:
Actully, if you open the Code Quality Check and open `Step Two - Check Lint Errors` you can see all of the warnings

oleonardolima (commented):
@benalleng I think we could have it as part of the pre-commit command as well, wdyt ?

benalleng:
since its an eslint it should just run there, the pre-commit command runs eslint already `"pre-commit": "pnpm run eslint && pnpm run check-types && pnpm run format",`

benalleng:
Upon thinking more about When this command should be run I decided to make a new CI command because I don't think that we need to get 800+ warnings on every pre-commit and I don't think translators need to be modifying eslint rules to check their work so I created a `check-i18n-keys` command

package.json
```diff
@@ -7,8 +7,9 @@
         "dev": "vite",
         "build": "tsc && vite build",
         "check-types": "tsc --noemit",
-        "eslint": "eslint src",
```
futurepaul, line 10: this name-change is causing the code check failure

futurepaul (commented):
I've never written an eslint plugin and today is def not the day I want to start, but if this works and doesn't interfere with the regular workflow (I appreciate you moving this to an optional command that can be run when needed) I'm inclined to merge. Just need to confirm it's working!

package.json
benalleng: I had to do this because the justfile doesn't like the shared name between eslint the bash command and eslint the pnpm script, i fixed it though, had to add a new eslint-path for a "bare" eslint command

tools/internal-rules/package.json
benalleng, line 1: should I move this eslint plugin to the `/scripts` directory?

futurepaul: nah tools is fine with me. can you add something to the readme? otherwise this is good to go works on my machine!

benalleng: Went ahead and added a full translation part for the README, ive seen it sometimes as a completely separate README but up to you where you want it
https://github.com/MutinyWallet/mutiny-web/blob/8d9af5e3a901a0e972da877cdb7dcb389dc59599/README.md#translating

futurepaul (approved):
thank you! and the readme is amazing this will be so helpful

futurepaul (approved):

PR #893, opened 2024-02-20
Fix identifiying zap kind

benthecarman:
Before we could never properly identify anon zaps because the `anon` tag we were getting from `findByTag` would be `undefined`  because there was no second element, even though it did have the `anon` tag. This just makes a dedicated function so it is easier to reason about.


futurepaul (approved):
I already added this to my social branch but can't hurt to merge here as well it's working for me

PR #899, opened 2024-02-22
Merge waila

futurepaul:
needs https://github.com/MutinyWallet/mutiny-node/pull/1045

### DIFF
--- a/src/logic/waila.ts
+++ b/src/logic/waila.ts
@@ -1,9 +1,6 @@
-import initWaila, { PaymentParams } from "@mutinywallet/waila-wasm";
+import { PaymentParams } from "@mutinywallet/mutiny-wasm";
 
 import { Result } from "~/utils";
-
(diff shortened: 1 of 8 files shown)


src/logic/waila.ts
```diff
@@ -1,10 +1,7 @@
-import initWaila, { PaymentParams } from "@mutinywallet/waila-wasm";
+import { PaymentParams } from "@mutinywallet/mutiny-wasm";
 
 import { Result } from "~/utils";
 
-// Make sure we've initialzied waila before we try to use it
-await initWaila();
```
benthecarman, line 6: Is there any concern with mutiny-wasm not being initialized yet?

futurepaul: no, if we try to use waila before mutiny-wasm is initialized we have bigger problems

benthecarman (approved):

PR #908, opened 2024-02-27
set lsps settings via url query params

johncantrell97:
Using this for internal testing atm but would be nice to have _something_ like this in mutiny. I doubt this is how it really should be implemented so open to feedback on the approach.

It works fine for me, for what it's worth.

Basically it does the same thing the form submission does on the LSPS section of /admin/settings and then instead of refreshing the page it navigates to wherever the user was headed.


futurepaul:
closing in favor of #1001

PR #917, opened 2024-03-07
refactor translations for easier editing

futurepaul:
I need to update the readme, but I wanted better tooling for translations and it seemed the easiest way to get this working was with .json files

also by moving the translations to the public folder I think I fixed the http backend stuff?

cool upside is the i18n-ally editor works now:

Screenshot 2024-03-07 at 3 41 10 PM

also i18n json editor allows for bulk editing:

Screenshot 2024-03-07 at 3 42 22 PM

cc @benalleng thoughts?


benalleng (commented):
Wow! This makes it much nicer moving forward! Looks great! incidentally this also makes us able to use the goDaddy i18n-eslint plugin with this [rule](https://github.com/godaddy/eslint-plugin-i18n-json?tab=readme-ov-file#i18n-jsonidentical-keys) to replace my eslint creation from #892 if we even need it anymore wth these bulk edit tools

benalleng:
Gave it a test on a fork and the i18n-json plugin works nicely, we definitely should get rid of the `internal-rules` as they really are designed for checking against 2 typescript objects

futurepaul:
Need to get this in for my own quality of life, for some reason recently typescript is choking on all the translation stuff and it breaks other parts of my editor. Updated the readme and removed the internal-rules stuff. Would love to get an opt-in checking flow as a followup with something like `i18n-eslint` if you're down to do that @benalleng but I can live without it for now.

benalleng (approved):
Looks good to me, always nice to see some negative diff.  Happy to add in the i18n-eslint later

PR #918, opened 2024-03-08
Mint discoverability

benthecarman:
for https://github.com/MutinyWallet/mutiny-node/pull/1060

works with my events i created, I don't think this will handle well if information is missing or if there is a ton of people recommending 

also need to do i18n and stuff too

image

### DIFF
--- a/src/routes/settings/ManageFederations.tsx
+++ b/src/routes/settings/ManageFederations.tsx
@@ -5,7 +5,7 @@ import {
     setValue,
     SubmitHandler
 } from "@modular-forms/solid";
-import { FederationBalance } from "@mutinywallet/mutiny-wasm";
+import { FederationBalance, TagItem } from "@mutinywallet/mutiny-wasm";
 import { A, useSearchParams } from "@solidjs/router";
 import { Scan } from "lucide-solid";
 import {
@@ -30,7 +30,10 @@ import {
     FancyCard,
     InfoBox,
     KeyValue,
+    LabelCircle,
     LargeHeader,
+    LoadingSpinner,
+    MediumHeader,
     MiniStringShower,
     MutinyWalletGuard,
     NavBar,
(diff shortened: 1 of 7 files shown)


futurepaul:
I rebased and spruced it up a little aesthetically. right now I'm not seeing any recommends and it doesn't seem like my recommend shows up? but it's a super slick ux for adding!

benthecarman:
>  I'm not seeing any recommends and it doesn't seem like my recommend shows up? but it's a super slick ux for adding!

i don't see any created my your npub

futurepaul:
sweet idk if my other wallet was in a bad state nsec-wise, but this worked:

Screenshot 2024-03-25 at 4 27 34 PM

question: is there a good way to check if we've already recommended a fedimint so we can hide the recommend button?

benthecarman:
> question: is there a good way to check if we've already recommended a fedimint so we can hide the recommend button?

hmm, not at the moment, could add some stuff for that. Maybe for now just save in local storage

src/routes/settings/ManageFederations.tsx
```diff
@@ -107,50 +143,229 @@ function AddFederationForm(props: { refetch?: RefetchType }) {
...
+                const myRecommend = thisFederation.recommendations.find(
+                    (r) => r.npub === myHexpub
+                );
+                return myRecommend;
+            }
+        } catch (e) {
+            console.error(e);
+            return undefined;
+        }
+    });
```
benthecarman, line 334: I'll add a function to mutiny-node that does this more efficiently, but shouldn't be blocking

benthecarman:
can't approve but lgtm

futurepaul:
updated to use https://github.com/MutinyWallet/mutiny-node/pull/1086

TonyGiorgio:
Build is failing, can't test atm

TonyGiorgio:
Screenshot 2024-03-26 at 7 31 27 PM

I'm running on mainnet, this shouldn't be happening.

TonyGiorgio:
This should be fixed with the next RC release. 

In the meantime, can we remove the invite code here? Kind of unnecessary, and when there's 4 or more, it takes up a lot of room and is even more unnecessary.

TonyGiorgio (approved):

PR #921, opened 2024-03-12
Fix NIP-05 handling for zap feed

benthecarman:
We support entering a nip-05 address in the sync nostr contacts form, however this would break the zap feed because it'd always return `undefined` when trying to get the hex pub. This fixes  it so we can actually resolve it into the hexpub as well as makes it so it'll return a better error when this happens.


benthecarman:
not needed anymore

PR #927, opened 2024-03-20
Add i18n-eslint json

benalleng:
This adds the eslint for json for translators to add some opt-in checking flow for translators.

At the moment it adds a key ordering warning, but I don't know that we ever created a "proper" ordering for our keys so I'm not sure if we wanted to "--fix" those


benalleng:
Stale. It was fun working on Mutiny!

PR #929, opened 2024-03-21
Add error message for tor urls

benthecarman:
image

Some people keep trying this and borking their wallet. This will give them an error for tor support


benthecarman:
This might break some of the start9 stuff, need to verify

TonyGiorgio (approved):

PR #932, opened 2024-03-22
Always pass along lnurl comment

benthecarman:
We were only doing this for zaps but we can pass it along for lnurls too


TonyGiorgio:
Needs rebase

TonyGiorgio (approved):

Issue #935, opened 2024-03-23
QR scanner not working with iPhone 15


benthecarman:
A user reported to me on twitter, don't have an iphone 15 so not sure how to reproduce.

This is in the PWA

futurepaul:
Haven't seen this from anyone else so gonna close

PR #936, opened 2024-03-23
Set inversion mode for scanner to 'both'

benthecarman:
Was reading through the docs and seems like this would be beneficial 

https://github.com/nimiq/qr-scanner?tab=readme-ov-file#color-inverted-mode

potentially related to the user's issue in #935


futurepaul (approved):
inverted qr is the first don't on this list lol https://www.qrcode-monkey.com/6-reasons-why-your-qr-code-is-not-working/

PR #938, opened 2024-03-25
Fix anonymous zaps

benthecarman:
https://github.com/MutinyWallet/mutiny-node/pull/1040 broke anon zaps, now requires you to put the type of zap you're doing.

We should add public and private zaps eventually, but this fixes the functionality for now


futurepaul (approved):
I don't have a good way to test but I trust you

benthecarman:
> I don't have a good way to test 

you can zap the faucet

PR #939, opened 2024-03-25
Better profile editing and clear

futurepaul:
Two unrelated commits sry but the npx sync just updates some native deps and the profile editing stuff is a pretty bad bug makes it easier to test quickly


benthecarman (approved):
Will this fix the ios warnings we've been getting?

futurepaul:
No but I think that's a pretty simple update as well can do that soon

PR #940, opened 2024-03-25
update to cap 6

### LINKED ISSUE #887: Target iOS 17+
```
ITMS-90725: SDK version issue - This app was built with the iOS 16.2 SDK. Starting April 29, 2024, all iOS and iPadOS apps must be built with the iOS 17 SDK or later, included in Xcode 15 or later, in order to be uploaded to App Store Connect or submitted for distribution. 
```

futurepaul:
cap 5's ios sdk version is EOL end of april, cap 6 is still in RC but seems pretty stable

will need to test this on devices before merge

(everything here is an auto migrate other than the api change for the global listeners)

closes #887


futurepaul:
Okay so everything is now with a "real" 6.0.0 release version except the secure-storage-plugin hasn't been updated and still has a peer dependency of 5.0... we could test if it still works, or we could look for an alternative

futurepaul:
this happened in #1091

PR #949, opened 2024-03-26
Remove tos

TonyGiorgio:


src/routes/setup/Root.tsx
```diff
@@ -67,12 +67,6 @@ export function Setup() {
                         Import Existing
                     </Button>
                 </div>
-                <div class="flex-1" />
-                <p class="max-w-[15rem] text-center text-xs font-light text-m-grey-400">
-                    By creating or restoring a wallet you agree to our{" "}
-                    <a href="#">Terms of Service</a>.
-                </p>
-                <div class="flex-1" />
```
futurepaul, line 75: need to keep one of these `<div class="flex-1" />` divs for layout reasons

futurepaul (changes requested):

futurepaul (approved):

PR #951, opened 2024-03-26
Remove local storage npub

benthecarman:
this will fix the zap feed

closes https://github.com/MutinyWallet/mutiny-node/issues/1088

### DIFF
--- a/src/components/NostrActivity.tsx
+++ b/src/components/NostrActivity.tsx
@@ -40,7 +40,10 @@ export function NostrActivity() {
     const i18n = useI18n();
     const [state, _actions] = useMegaStore();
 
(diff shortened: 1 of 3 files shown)


futurepaul:
this doesn't work but I can fix

benthecarman:
> this doesn't work but I can fix

i ran `just pre` and everything passed. Will never understand what typescript actually enforces

futurepaul (approved):
sry it was just that the call didn't have a `()` on the end of it, not a typescript formatting thing. I figured there would be a bunch more places where we used `state.npub` so I wanted to go over it, but happy to find that not the case!

PR #959, opened 2024-03-27
200k sat warning

### LINKED ISSUE #953: Explain 200k sat federation limit
Not sure where we should put this, but the number one question I get from people that have recently added a federation is "how do I receive to it?"

I think either on the federation page or on the QR receive page, we should indicate something along the lines of: 

"receives under 200k go into your federation by default, anything more than that goes into your lightning channel"

I'm leaning more towards the receive screen? Just because this number might increase or decrease at a later time and the user should be aware of it.

futurepaul:
closes #953

### DIFF
--- a/src/routes/Receive.tsx
+++ b/src/routes/Receive.tsx
@@ -3,7 +3,7 @@ import {
     MutinyInvoice
 } from "@mutinywallet/mutiny-wasm";
 import { useNavigate } from "@solidjs/router";
-import { ArrowLeftRight } from "lucide-solid";
+import { ArrowLeftRight, CircleHelp, Users, Zap } from "lucide-solid";
 import {
     createEffect,
     createMemo,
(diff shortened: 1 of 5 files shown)


src/routes/Receive.tsx
```diff
@@ -102,6 +102,33 @@ function FeeWarning(props: { fee: bigint; flavor: ReceiveFlavor }) {
...
 }
 
+function ReceiveMethodHelp(props: { amountSats: bigint }) {
+    const i18n = useI18n();
+    const [open, setOpen] = createSignal(false);
+    return (
+        <>
+            <button class="flex gap-2 self-end" onClick={() => setOpen(true)}>
+                <Switch>
+                    <Match when={props.amountSats < 200000n}>
```
TonyGiorgio, line 112:
```suggestion
                    <Match when={props.amountSats <= 200000n}>
```

TonyGiorgio (approved):
Works great!

PR #961, opened 2024-03-27
Add `lightning:` prefix to ln addr qr code

benthecarman:
Without then when I scan the qr it tries to send an email


futurepaul (approved):
nice catch

PR #962, opened 2024-03-27
Last UI fixes

futurepaul:
I had intended to have this different route for nostr import after setup (that's why the form was already its own component), it's pretty confusing as-is.


benthecarman:
there's no back button or any way to exit the import profile screen

benthecarman (approved):

PR #965, opened 2024-03-27
Handle contacts with no npub/lnurl

### LINKED ISSUE #963: Handle contacts with no npub or lightning address
We currently have contacts in our social bar that if you click on them it'll take you to a black screen if they don't have an npub.

We should handle this by taking you to the send screen if they have a lightning address or just not displaying them if they don't have either

benthecarman:
Closes #963 

Not really sure if `actions.handleIncomingString` is the correct thing to be calling here

### DIFF
--- a/src/components/SocialActionRow.tsx
+++ b/src/components/SocialActionRow.tsx
@@ -15,8 +15,16 @@ export function SocialActionRow(props: {
 
     const getContacts = cache(async () => {
         try {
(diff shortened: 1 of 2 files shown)


futurepaul (commented):
this looks good to me. it was scary at first but it makes sense on re-reading.

should get it in after we merge #989 and we can test that it all works together

futurepaul (approved):

PR #970, opened 2024-03-28
add delete nostr profile ability

futurepaul:
needs https://github.com/MutinyWallet/mutiny-node/pull/1096

### DIFF
--- a/src/routes/Profile.tsx
+++ b/src/routes/Profile.tsx
@@ -1,5 +1,5 @@
 import { useNavigate } from "@solidjs/router";
-import { Copy, Edit, QrCode } from "lucide-solid";
+import { Copy, Edit, Import, QrCode } from "lucide-solid";
 import { createMemo, createSignal, Match, Show, Switch } from "solid-js";
 import { QRCodeSVG } from "solid-qr-code";
 
@@ -33,7 +33,8 @@ export function Profile() {
         return {
             name: profile?.display_name || profile?.name || "Anon",
             picture: profile?.picture || undefined,
-            lud16: profile?.lud16 || undefined
+            lud16: profile?.lud16 || undefined,
+            deleted: profile?.deleted || false
         };
     });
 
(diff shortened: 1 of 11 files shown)


src/routes/settings/NostrKeys.tsx
```diff
@@ -18,12 +20,57 @@ import {
...
+    }
+
+    const [confirmOpen, setConfirmOpen] = createSignal(false);
+    const [confirmLoading, setConfirmLoading] = createSignal(false);
+
+    async function deleteNostrAccount() {
+        setConfirmLoading(true);
+        try {
+            await state.mutiny_wallet?.delete_profile();
+            console.log("yooo");
```
benthecarman, line 38: don't need tihs console log

src/routes/settings/NostrKeys.tsx
```diff
@@ -18,12 +20,57 @@ import {
...
+        setConfirmOpen(true);
+    }
+
+    const [confirmOpen, setConfirmOpen] = createSignal(false);
+    const [confirmLoading, setConfirmLoading] = createSignal(false);
+
+    async function deleteNostrAccount() {
+        setConfirmLoading(true);
+        try {
+            await state.mutiny_wallet?.delete_profile();
```
benthecarman, line 37: should delete the local nsec if one is imported too

benthecarman (approved):
works for me

TonyGiorgio (changes requested):
Importing an nsec after deleting the main wallet nsec doesn't work.

futurepaul:
> Importing an nsec after deleting the main wallet nsec doesn't work.

It works for me. I made a new profile with a new wallet, deleted the profile, and then imported my normal nsec.

src/routes/Profile.tsx
```diff
@@ -42,11 +42,21 @@ export function Profile() {
...
     const [copy, copied] = useCopy({ copiedTimeout: 1000 });
 
+    const profileDeleted = createMemo(() => {
+        // return true if profile.deleted is true or "true"
+        // return false if profile.deleted is false or "false"
+        if (profile().deleted === true || profile().deleted === "true") {
+            return true;
+        } else {
+            return false;
+        }
```
benthecarman, line 52: this can just be `(profile().deleted === true || profile().deleted === "true")` no need for the if statement

benthecarman (approved):
Paul and I tested every case we could think of on a call, this is good to go

TonyGiorgio (approved):

TonyGiorgio (approved):

PR #976, opened 2024-03-29
i18n for `prettyPrintTime` functions

alpeb-btc:
`openLinkProgrammatically.ts` mentions that i18n doesn't work in utils; maybe that comment is outdated?

I've also added the translations for Spanish and that tested well.

Removed the `formatExpiration` function as it appears to no longer be used.


futurepaul:
> `openLinkProgrammatically.ts` mentions that i18n doesn't work in utils; maybe that comment is outdated?

yeah I'm surprised this works but it all seems fine for me!

thank you for tackling this, good stuff

futurepaul (approved):

PR #1001, opened 2024-04-05
set lsps settings via url query params

futurepaul:
sorry @johncantrell97 it was easier to do this than rebase / change your version. works on my machine but if you can test as well that would be great and we can get this in


benthecarman (approved):
Works for me

PR #1020, opened 2024-04-09
Add italian translations

TonyGiorgio:


futurepaul (changes requested):
Need to add the language to https://github.com/MutinyWallet/mutiny-web/blob/master/src/utils/languages.ts so it shows up in the language picker dropdown

PR #1030, opened 2024-04-10
Skip setup if flag set & set LSPS Config on startup

benthecarman:
if you add `?skip_setup=true` to the url it'll now skip our walkthrough, this is for the c= guys


futurepaul (commented):
doesn't this mean they won't end up with a nodemanager?

benthecarman:
> doesn't this mean they won't end up with a nodemanager?

No, this just skips the setup profile stuff

futurepaul (approved):

TonyGiorgio (approved):

PR #1035, opened 2024-04-11
pick a federation on setup

futurepaul:
Screenshot 2024-04-11 at 9 55 33 AM

### DIFF
--- a/e2e/utils.ts
+++ b/e2e/utils.ts
@@ -13,10 +13,14 @@ export async function loadHome(page: Page) {
 
     await page.locator("button:has-text('Skip for now')").click();
 
+    await page.getByText("Pick a Federation").waitFor();
+
+    await page.locator("button:has-text('Skip for now')").click();
+
+    await page.locator(`button:has-text('Confirm')`).click();
+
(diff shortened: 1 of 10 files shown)


TonyGiorgio:
Amazing, I love it. I think the copy could improve, both here and in the setup. Been asked a few times what this means. I think we can explain it a bit better, something that focuses on what it provides the user, and something a little less scary than "use at your own risk" while still communicating that each is a new network. Hard to wrap it all up in a small package, could benefit from a "learn more" screen itself.

futurepaul:
> I think the copy could improve, both here and in the setup. 

I'll take another pass at it. Really hard to explain without a wall of text but I'll try again. I can add a learn more explainer modal for the truly curious.

futurepaul:
@TonyGiorgio updated with the text you provided

TonyGiorgio (approved):
LFG

benthecarman:
think this broke the playwright test

futurepaul:
> think this broke the playwright test

it was the confirm button's fault

PR #885, opened 2024-02-15
redeem cashu token

elnosh:
Still a WIP but putting it out there for feedback. This PR is to add page to redeem Cashu token.

It depends on:
- mutiny-node PR: https://github.com/MutinyWallet/mutiny-node/pull/1026
- waila change that added support for parsing Cashu token: https://github.com/MutinyWallet/bitcoin-waila/pull/17

So right now it does not work as is. I'm building both changes mentioned above locally and pointing to them in package.json file.

Couple things still need work/fixing:
- CORS error when calling mint.
- After pressing Redeem button, UI is not responsive because of CORS error. It does however add it to the wallet balance if the mint pays the invoice.
- I was trying cashu tokens generated from different wallets and was getting some errors when parsing some of them (I think this may need changes in Waila)

PR #1038, opened 2024-04-11
Cashu redeem

futurepaul:
We can close #885 and go with this

I added string handling, redeem screen stuff, and mild support in DMs. in the future we can do contact tagging and auto-sweep for DMs, but this seemed easy enough for now.


src/routes/Redeem.tsx
```diff
@@ -238,7 +310,23 @@ export function Redeem() {
                             </div>
                             {/* TODO: add payment details */}
                         </SuccessModal>
-                        <pre>NICE</pre>
```
benthecarman, line 241: why delete this nice?

futurepaul: idk what it's even doing there tbh, it should be behind the successmodal anyway

benthecarman (approved):

futurepaul:
Works on signet but seems like there are some problems redeeming with real mainnet mints so switching to draft until we figure that out

benthecarman:
requires https://github.com/MutinyWallet/mutiny-node/pull/1139

benthecarman:
this should work with the latest release

PR #1043, opened 2024-04-12
invite codes and list federations even if you're in one

futurepaul:
@TonyGiorgio and I discussed this over steak tartare

Screenshot 2024-04-12 at 11 41 17 AM

Screenshot 2024-04-12 at 11 41 21 AM

### DIFF
--- a/src/components/FederationInviteShower.tsx
+++ b/src/components/FederationInviteShower.tsx
@@ -1,0 +1,55 @@
+import { Copy, QrCode } from "lucide-solid";
+import { createSignal, Show } from "solid-js";
+import { QRCodeSVG } from "solid-qr-code";
+
+import { SimpleDialog } from "~/components";
+import { useI18n } from "~/i18n/context";
+import { useCopy } from "~/utils";
+
+export function FederationInviteShower(props: {
+    name?: string;
+    inviteCode: string;
+}) {
+    const i18n = useI18n();
+    const [showQr, setShowQr] = createSignal(false);
+
+    const [copy, copied] = useCopy({ copiedTimeout: 1000 });
+
+    return (
+        <>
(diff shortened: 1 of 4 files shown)


benthecarman:
If you guys want this that is fine with me, but curious what is the reasoning for this?

src/components/FederationInviteShower.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <div class="flex w-full justify-center gap-8">
+                <button onClick={() => setShowQr(true)}>
+                    <QrCode class="inline-block h-4 w-4" />
+                </button>
+                <button
+                    class="p-1"
+                    classList={{
+                        "bg-m-red rounded": copied()
+                    }}
+                    onClick={() => copy(props.inviteCodes[0])}
```
benthecarman, line 29: Would be cool if you could scroll through the invite codes, in case the first isn't working

futurepaul:
> If you guys want this that is fine with me, but curious what is the reasoning for this?

it's just nice for reasons of curiosity imo

src/components/FederationInviteShower.tsx
futurepaul: we don't actually have multiple. made it singular

benthecarman (commented):
when you remove a federation it reloads the list still, shoulnd't need to do that anymore but not the end of the world

futurepaul:
> when you remove a federation it reloads the list still

yeah it's a lot simpler code-wise this way

benthecarman (approved):

TonyGiorgio:
> If you guys want this that is fine with me, but curious what is the reasoning for this?

multiple people are asking me for it, they want to know the federations that exist.

PR #1046, opened 2024-04-12
handle invoices inside of message

futurepaul:
fixes unleashed.chat @benthecarman 

Screenshot 2024-04-12 at 2 53 59 PM

Screenshot 2024-04-12 at 2 54 08 PM

### DIFF
--- a/src/routes/Chat.tsx
+++ b/src/routes/Chat.tsx
@@ -76,9 +76,20 @@ function SingleMessage(props: {
 
     const parsed = createAsync(
         async () => {
-            const result = toParsedParams(props.dm.message, network);
+            let result = undefined;
 
-            if (!result.ok) {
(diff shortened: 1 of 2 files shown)


src/routes/Chat.tsx
```diff
@@ -92,6 +103,10 @@ function SingleMessage(props: {
                         return {
                             type: "invoice",
                             status: "paid",
+                            original_message: props.dm.message.replace(
```
benthecarman, line 106: can you call this `message_without_invoice` or something. `original_message` feels like the whole message and will be confusing when we haven't looked at this code in 6 months

futurepaul: good call

benthecarman (approved):
awesome!

PR #1074, opened 2024-04-18
Remove swap functionality for now

TonyGiorgio:
This has caused nothing but problems we can't reproduce, more than likely LSP related problems anyways.


src/components/BalanceBox.tsx
futurepaul, line 1: just need to delete these unused bits and looks good to me. I agree it's a large source of pain and suffering

futurepaul: and for the record, if you want to comment something out in the jsx section:
```
{/* <Show when={usableOnchain() > 0n}>
    <div class="self-end justify-self-end">
        <A href="/swap" class={STYLE}>
            <Shuffle class="h-6 w-6" />
        </A>
    </div>
</Show> */}
```

TonyGiorgio: retarded

TonyGiorgio:
Can you approve this now @futurepaul ?

PR #1078, opened 2024-04-18
handle no connection

futurepaul:
if they go offline and the device can detect it we'll show them they're offline

https://capacitorjs.com/docs/apis/network

Screenshot 2024-04-18 at 4 15 24 PM


src/router.tsx
```diff
@@ -125,6 +126,9 @@ function ChildrenOrError(props: { children: JSX.Element }) {
                     password={state.password}
                 />
             </Match>
+            <Match when={state.network_status?.connectionType === "none"}>
+                <NoConnection />
+            </Match>
```
benthecarman, line 131: this should probably be before `setup_error`. having no internet would likely be the cause of a setup error

TonyGiorgio: One of the first things that gets returned from the setup phase is the LNURL call not working

futurepaul: I'm not sure if you guys are agreeing or disagreeing here.

benthecarman: agreeing

futurepaul: okay rebased and fixed

TonyGiorgio:
I don't like this. It's very jarring. It's one thing if it's during the setup, but if there's a little disconnection while on the home screen, the whole screen converts and flashes back when the connection resumes. This is not a good experience for a blip in connection issues. You can do a little disconnection icon at the top or something.

PR #1082, opened 2024-04-19
bump the timeout and make sure we're not done loading

futurepaul:
Someone reported the wallet was loading and then hitting the timeout thing after. I don't know how that would happen, it would seem somehow the timeout interval isn't being cleared. This makes it so even if the timeout is triggered, it won't do anything if the loading is done.

### DIFF
--- a/e2e/roundtrip.spec.ts
+++ b/e2e/roundtrip.spec.ts
@@ -124,6 +124,9 @@ test("rountrip receive and send", async ({ page }) => {
 
     await page.click("text=Online Channels");
 
+    // Give it just a second to settle down
+    await page.waitForTimeout(2000);
+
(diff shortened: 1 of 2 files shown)


TonyGiorgio:
Did you test it with a password? Because that's what they said.

futurepaul:
Yeah I tested with a password. I think that was the problem... if they need to set a password it throws out of this function before the interval gets cleared. Now it checks if setup is done, so as long as they enter their password under 90s they're good.

TonyGiorgio:
Okay, well your original statement said you didn't know how it could happen. Glad you were able to reproduce and fix it.

TonyGiorgio (approved):

PR #1089, opened 2024-04-22
get rid of unified receive

futurepaul:
No more bip21, just lightning and bitcoin address. need to test this out with ben's upcoming pr so we make sure we can handle all the potential scenarios

### DIFF
--- a/e2e/fedimint.spec.ts
+++ b/e2e/fedimint.spec.ts
@@ -77,10 +77,10 @@ test("fedmint join, receive, send", async ({ page }) => {
 
     const value = await qrCode.getAttribute("value");
 
-    // The SVG's value property includes "bitcoin:t"
-    expect(value).toContain("bitcoin:t");
+    // The SVG's value property includes "bitcoin:l"
+    expect(value).toContain("lightning:l");
 
-    const lightningInvoice = value?.split("lightning=")[1];
+    const lightningInvoice = value?.split("lightning:")[1];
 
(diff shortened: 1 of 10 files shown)


futurepaul:
https://no-more-unified.mutiny-web.pages.dev/

TonyGiorgio (changes requested):
The API calls need to be different. You're still calling for a bip21 address, as indicated by the logs that are checking for an address even when the lightning invoice is the only thing present.

TonyGiorgio:
Screenshot 2024-05-08 at 1 17 19 PM

This is incorrect.

TonyGiorgio:
Also, we should allow someone to just hit continue for 0 amount on chain. There's no reason they need to do anything for it.

futurepaul:
addressed @TonyGiorgio comments and also @benthecarman I think I fixed the flicker #1125

futurepaul:
closes #1126 as well

TonyGiorgio:
I also fix dust amount errors in my fedimint onchain mutiny-node PR, but good to have it in both.

TonyGiorgio (approved):
Works great! Only nit, why not auto select and minimize once onchain or lightning is clicked? We did that before and I liked it. Not sure if you felt one way or another, but that's the only thing that's weird to me.

futurepaul:
Good idea, i think i had to get rid of that when we had the remember selection checkbox

PR #1091, opened 2024-04-23
web worker

### LINKED ISSUE #601: Service worker
## goal:

if we run mutiny inside a service worker we get a lot of benefits:
* improved latency (don't block the main thread)
* able to handle push notifications (potentially handle receive when the app isn't open)
* better caching (this is possible now but we need a better "bundle" story)

other potential stuff it unlocks:
* fine-grained control over updates
* verifying signature / hash of new versions

## challenges:
will need to wrap all of mutiny-node's APIs in an event-based or route-based API so it's callable from the frontend

## research:

chrome://serviceworker-internals/

Resources:
https://news.ycombinator.com/item?id=33228891
https://github.com/richardanaya/wasm-service/blob/main/sw.js#L19
https://github.com/elk-zone/elk/blob/main/service-worker/sw.ts

futurepaul:
closes #601 (kind of)

I've confused the concept of ServiceWorker and SharedWorker a lot of times. What we have here is a [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker)... in the future this will make it easier for a ServiceWorker to connect to the one canonical instance of Mutiny, whether or not the app is open.

For now, the main win is getting all the mutiny work off the main UI thread. Everything should be more responsive (no blocking) and in some cases faster (no contention, so faster, though there is a serialization cost going across threads).

EDIT:
it's a web worker now, not a shared worker. thx chrome android

TODO:
- [x] add the doc strings
- [x] test everything


benthecarman:
in some places you do `sw.function()` and others you do `sw?.function()` what is the reasoning there?

src/router.tsx
```diff
@@ -172,14 +168,11 @@ export function Router() {
...
             <Route path="/chat/:id" component={Chat} />
             <Route path="/feedback" component={Feedback} />
-            <Route path="/gift" component={GiftReceive} />
             <Route path="/receive" component={Receive} />
             <Route path="/redeem" component={Redeem} />
             <Route path="/request/:id" component={RequestRoute} />
             <Route path="/scanner" component={Scanner} />
             <Route path="/send" component={Send} />
-            <Route path="/swap" component={Swap} />
-            <Route path="/swaplightning" component={SwapLightning} />
```
benthecarman, line 182: Think we're only temporarily disabling these for now, should bring them back eventually

src/routes/Chat.tsx
```diff
@@ -447,26 +449,18 @@ function FixedChatHeader(props: {
...
-    //         console.error(e);
-    //         return undefined;
-    //     }
-    // });
-
     const [contact, { refetch: refetchContact }] = createResource(async () => {
         try {
-            return state.mutiny_wallet?.get_tag_item(params.id);
+            const item = await sw.get_tag_item(params.id);
+            console.log("item", item);
```
benthecarman, line 462: this will spam the logs

src/routes/Send.tsx
```diff
@@ -235,7 +240,7 @@ export function Send() {
 
     // send?invoice=... need to check for wallet because we can't parse until we have the wallet
     createEffect(() => {
-        if (params.invoice && state.mutiny_wallet) {
+        if (params.invoice && !state.wallet_loading) {
```
benthecarman, line 243: in other places you were checking the `load_stage === "done"`

benthecarman (commented):
Everywhere where you do `// PAIN` in the worker is where we return a wasm_bindgen type instead of a `JsValue` this was supposed to be nicer for you but seems like it is worse now. Maybe in mutiny-wasm we should just try to do `JsValue` everywhere now

benthecarman:
> What we have here is a [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker).

this site says it doesn't work for chrome android :cry:, will need to test

futurepaul:
> > What we have here is a [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker).
> 
> this site says it doesn't work for chrome android 😢, will need to test
> 
> probably should add to the screen where we detect if js/wasm is enabled, that shared workers are available

THIS IS SO SAD :(. It's a one line change to make it a non-shared worker, but I might look into detecting Chrome Android specifically so we can use sharedworker other places. But that means I need to bring back all the double init defense stuff :(

I'm also having trouble with capacitor right now. It's not working iOS or Android, for what might be different reasons. One problem is this `import.meta.url` which doesn't seem to exist in the Capacitor environment, trying to figure it out.

futurepaul:
Got android working...

Screenshot 2024-04-24 at 11 35 12 AM

But iOS is mad about something. Works in desktop safari...

Screenshot 2024-04-24 at 11 37 07 AM

futurepaul:
> Everywhere where you do `// PAIN` in the worker is where we return a wasm_bindgen type instead of a `JsValue` this was supposed to be nicer for you but seems like it is worse now. Maybe in mutiny-wasm we should just try to do `JsValue` everywhere now

yeah it sucks because the JsValue erases the type so I have to go digging to figure out what it's actually returning. do you remember why this was rejected btw? https://github.com/MutinyWallet/mutiny-node/pull/882

benthecarman:
Wasm bindgen does the same thing, it just doesn't work for async functions yet. I think they recently merged support for it though

TonyGiorgio:
My mainnet tests on my android phone are great. No more flicker. Page switches are much faster now. Signet will need a full end to end test but this is really smooth. Nice work!

TonyGiorgio:
Screenshot 2024-05-02 at 4 34 57 PM

I don't think it's intended to be at the bottom?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "mutiny-wallet",
-    "version": "0.6.8",
+    "version": "0.6.9",
```
TonyGiorgio, line 3: regenerate package.lock

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #1100, opened 2024-04-27
restore swap functionality

futurepaul:
reverts #1074


benthecarman (approved):
looks like we need a fix for ios builds, not blocking this though

futurepaul:
> looks like we need a fix for ios builds, not blocking this though

fixed in #1101 

something changed recently with the mac env and dumb ruby version stuff, thought it was related to cap 6 but was able to bring my changes from there to here and looks good now

PR #1101, opened 2024-04-27
remove 200k language

futurepaul:
we're removing the 200k limit, now the info box just lets them know it will go into a federation


benthecarman (approved):

futurepaul:
something changed recently with the mac env and dumb ruby version stuff, thought it was related to cap 6 but was able to bring my changes from there to here and looks good now

PR #1106, opened 2024-04-28
Force close channel if offline

benthecarman:
Had to help a user today where they were trying to close a channel while their channel was offline which always failed because it would try to cooperative close. This will make it force close if the channel is offline.


TonyGiorgio (approved):
concept ack

futurepaul (approved):
can't figure out how to keep the channel offline to test a force close but it looks good to me and everything else works

PR #1110, opened 2024-04-29
Correct grammar of federations description

j0hncc:
Correct "each one is ran by.. " to "each one is run by"


benthecarman:
these both seem right to me

futurepaul (approved):
pretty minor but I do agree "run" is more correct here as it's present tense

PR #1118, opened 2024-05-02
Consider fedimint balance for on chain sends

TonyGiorgio:
Depends on https://github.com/MutinyWallet/mutiny-node/pull/1146

TODO:
- [x] we should say not to send more than once to the address
- [x] we should tell them it requires 10 confs for deposits

### DIFF
--- a/src/routes/Receive.tsx
+++ b/src/routes/Receive.tsx
@@ -60,6 +60,7 @@ export type OnChainTx = {
         }>;
     };
     txid: string;
+    internal_id: string;
     received: number;
     sent: number;
     confirmation_time: {
@@ -213,7 +214,7 @@ export function Receive() {
         const paymentTxId =
             paidState() === "onchain_paid"
                 ? paymentTx()
-                    ? paymentTx()?.txid
+                    ? paymentTx()?.internal_id
                     : undefined
                 : paymentInvoice()
                   ? paymentInvoice()?.payment_hash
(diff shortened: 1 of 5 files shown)


src/routes/Send.tsx
```diff
@@ -825,6 +830,11 @@ export function Send() {
                             <p>{i18n.t("send.hodl_invoice_warning")}</p>
                         </InfoBox>
                     </Show>
+                    <Show when={source() === "onchain"}>
+                        <InfoBox accent="blue">
+                            {i18n.t("send.warn_address_reuse")}
+                        </InfoBox>
+                    </Show>
```
benthecarman, line 837: this needs to be on the receive side, not send side

futurepaul: okay thank you, fixed, moved to receive QR screen

src/routes/Receive.tsx
```diff
@@ -459,6 +460,16 @@ export function Receive() {
                                 <p>{error()}</p>
                             </InfoBox>
                         </Show>
+                        <Show
+                            when={
+                                flavor() === "onchain" &&
+                                state.federations?.length
```
benthecarman, line 466: we should just always show this for on-chain. we only sync an address once for our own on-chain wallet so we'll have a similar issue

futurepaul: fixed

benthecarman (approved):

PR #1127, opened 2024-05-07
Special case zapple pay for friends tab

### LINKED ISSUE #1122: Show zapplepay zaps correctly
This was supposed to solve it but it does not: 

https://github.com/MutinyWallet/mutiny-web/pull/1072

benthecarman:
Closes #1122

### DIFF
--- a/src/utils/fetchZaps.ts
+++ b/src/utils/fetchZaps.ts
@@ -8,6 +8,12 @@ import {
     NostrTag
 } from "~/utils/nostr";
 
+export const ZAPPLE_PAY_NPUB =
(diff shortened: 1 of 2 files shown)


src/utils/fetchZaps.ts
```diff
@@ -18,7 +24,7 @@ type NostrEvent = {
     sig?: string;
 };
 
-type SimpleZapItem = {
+export type SimpleZapItem = {
```
futurepaul, line 27: I don't think this needs to be exported?

futurepaul (approved):
nice, thank you. just one nit

Issue #1128, opened 2024-05-07
Popup info about user's federation


TonyGiorgio:
"popup_end_timestamp" and "popup_countdown_message" is what fedi uses for info. Needs the next release.

futurepaul:
I see the fields but they're all null except the meta_external_url (which looks like it has the data I need, I could just follow that for now and get the info here)

<img width="468" alt="Screenshot 2024-05-07 at 9 50 09 AM" src="https://github.com/MutinyWallet/mutiny-web/assets/543668/ad7f6b4b-19c3-427d-a623-a543a0bd7f98">

TonyGiorgio:
Okay, I might have to parse that manually and pull down the fields myself. I'll check with the fedimint team.

PR #1129, opened 2024-05-07
show federation expiration metadata

futurepaul:
for #1128 

here's what I have so far. putting this here so @TonyGiorgio can see the logic I used to get the metadata

Screenshot 2024-05-07 at 10 22 03 AM

### DIFF
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -23,14 +23,14 @@ export type MethodChoice = {
     maxAmountSats?: bigint;
 };
 
-// Make sure to update this when we get the fedi option in here!
 function methodToIcon(method: MethodChoice["method"]) {
-    if (method === "lightning") {
-        return "lightning";
-    } else if (method === "onchain") {
-        return "chain";
-    } else if (method === "fedimint") {
-        return "community";
+    switch (method) {
+        case "lightning":
+            return "lightning";
+        case "onchain":
+            return "chain";
+        case "fedimint":
+            return "community";
     }
 }
 
--- a/src/components/AmountEditable.tsx
+++ b/src/components/AmountEditable.tsx
@@ -19,7 +19,7 @@ import {
 } from "~/utils";
 
 export type MethodChoice = {
-    method: "lightning" | "onchain";
+    method: "lightning" | "onchain" | "fedimint";
     maxAmountSats?: bigint;
 };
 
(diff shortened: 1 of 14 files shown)


TonyGiorgio:
1. It's not obvious what I'm supposed to do after the "Manage Federations" pop up window comes up. I think if the "transfer funds" button stayed on the federation page no matter what, it would be more obvious. A popup could come up, saying whether they want to swap to lightning or swap to a new federation and they should add one if they have it?

2. When the federation is the 2nd one, the pop up doesn't happen.

futurepaul:
needs https://github.com/MutinyWallet/mutiny-node/pull/1169 and a new version of mutiny node

src/routes/Transfer.tsx
```diff
@@ -0,0 +1,220 @@
...
+    const [state, _actions, sw] = useMegaStore();
+    const i18n = useI18n();
+    const navigate = useNavigate();
+    const [amountSats, setAmountSats] = createSignal(0n);
+    const [loading, setLoading] = createSignal(false);
+    const [params] = useSearchParams();
+
+    const canTransfer = createMemo(() => {
+        return true;
+    });
```
benthecarman, line 41: this doesn't seem useful

src/routes/Transfer.tsx
```diff
@@ -0,0 +1,220 @@
...
+        } catch (e) {
+            const error = eify(e);
+            setTransferResult({ failure_reason: error.message });
+            console.error(e);
+        } finally {
+            setLoading(false);
+        }
+    }
+
+    // const fromFederatationId = params.from;
```
benthecarman, line 114: ping

TonyGiorgio (approved):

src/components/AmountEditable.tsx
```diff
@@ -29,6 +29,8 @@ function methodToIcon(method: MethodChoice["method"]) {
         return "lightning";
     } else if (method === "onchain") {
         return "chain";
+    } else if (method === "fedimint") {
+        return "community";
```
benthecarman, line 33: you have a comment here saying to add fedimint, you can remove that now.

also would be better as a switch than if statements

benthecarman (approved):

TonyGiorgio (changes requested):
Needs retest

futurepaul:
looks good now @TonyGiorgio

TonyGiorgio (approved):
Works great, just a nit that the federation name in the popup would be useful.

PR #1135, opened 2024-05-08
Filter self from social row

### LINKED ISSUE #1117: Own profile is auto-added as a contact
When I make a new wallet it generates a profile. After clicking around a bit (not exactly sure what triggers it) I end up with myself added as a contact. I've noticed this in the existing mutiny and in the new webworker version.

Screenshot 2024-05-02 at 2 27 56 PM

Screenshot 2024-05-02 at 2 30 30 PM

Seems like a @benthecarman problem

benthecarman:
Closes #1117


futurepaul (approved):
works great very filtered

PR #1145, opened 2024-05-09
Fix amountless onchain

### LINKED ISSUE #1143: Onchain address with no amount is not marking as payment initiated
With fedimint: 
- checks mempool but never updates (although I think the checks are what fedimint is doing)

Without fedimint:
- never checks mempool

benthecarman:
Closes #1143


futurepaul (approved):
ouch thank you

PR #1157, opened 2024-05-15
Add Chinese

ra5pvt1n:
Add translations for Simplified Chinese and Traditional Chinese. 

From README.md
> When creating a new language file ensure it follows the ISO 639 2-letter standard

Since scripts for both Simplified Chinese and Traditional Chinese share the same ISO 639 2-letter standard code (`zh`), I've provisionally named them as `hans.json` for Simplified Chinese and `hant.json` for Traditional Chinese in order to distinguish between them. 

A better way to name these files can be discussed. 

See also [How should the language codes be formatted?](https://www.i18next.com/how-to/faq#how-should-the-language-codes-be-formatted) from i18next FAQ


futurepaul (approved):
this is wonderful, thank you so much!

PR #1185, opened 2024-06-05
Start of sentry integration

TonyGiorgio:
- [x] TODO option to turn it on/off
- [x] Parse mutiny-node console and decide when to submit errors
- [x] ~~Find a better way to reference users/sessions than the default~~ just anon for now, removed the default
- [x] Source maps / github integration
- [x] Delete all testing data
- [x] Make sure any invoice/txid data is scrubbed client side


TonyGiorgio:
typescript checking is bogus, this is allowed: https://docs.sentry.io/platforms/javascript/configuration/filtering/#event-hints

TonyGiorgio:
Best to test with the latest with mutiny-node in order to get better results.

TonyGiorgio:
Now requires turning on in the admin tools

futurepaul (approved):
working well for me, finding panics and errors!

PR #1186, opened 2024-06-05
Move persistent storage init

benthecarman:
This was moved into the web worker which doesn't have access to this api call it seems


futurepaul (approved):
thank you

PR #1187, opened 2024-06-05
Move clear interval until after things are setup

benthecarman:
This fixed one version of setup getting stuck for me


futurepaul:
My worry is that we end up with more scenarios where the interval doesn't get cleared... really wish we had a better way of doing this timeout thing it's really not been fun.

benthecarman:
Made it so we'll clear it in the catch block as well, should be guaranteed to clear now

futurepaul (approved):
I thought I had a reason for not doing this in the catch block, but this seems to be working great and I can't think of why it can't be like this so let's go for it

TonyGiorgio (approved):

PR #1188, opened 2024-06-05
Immediately timestamp double init defense

benthecarman:
Before we would not set the session storage double init defense until after the wallet was initialized. This changes it to when we verify it isn't set, this way we immediately "claim the lock" instead of waiting until the wallet is loaded. Also cleaned up the error message to be more informative.

Not sure if this will actually fix anything but found this while trying to find init error


futurepaul (approved):
this is great, previously there were cases where the wallet would load and only then would it reload. I think this at least makes that loop a lot shorter.

PR #1195, opened 2024-06-07
default false diagnostics, style the welcome

futurepaul:
pushing to someone else's branch stresses me out, and also need some discussion here:

I'd prefer to not have this on the setup screen, but I'm willing to go with it if you feel strongly it should be there. However, I'm pretty strongly against it being default on.

This pr has some styling, a typo fix, and defaults to off.

### DIFF
--- a/src/routes/setup/Root.tsx
+++ b/src/routes/setup/Root.tsx
@@ -12,7 +12,7 @@ export function Setup() {
 
     const [isCreatingNewWallet, setIsCreatingNewWallet] = createSignal(false);
     const [isDiagnosticReportingEnabled, setIsDiagnosticReportingEnabled] =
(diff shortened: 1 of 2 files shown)


src/routes/setup/Root.tsx
```diff
@@ -12,7 +12,7 @@ export function Setup() {
 
     const [isCreatingNewWallet, setIsCreatingNewWallet] = createSignal(false);
     const [isDiagnosticReportingEnabled, setIsDiagnosticReportingEnabled] =
-        createSignal(true);
+        createSignal(false);
```
TonyGiorgio, line 15: Nack.

TonyGiorgio (changes requested):

PR #1205, opened 2024-06-11
Remove LSP url from servers

benthecarman:
First commit removes the LSP url from the servers page, this shouldn't be changed from here but from the admin page so we make sure we call the `change_lsp` function.

Second commit calls trim on all the lsp urls to help prevent issues


benthecarman:
Would be good to make the settings form just call `.trim` on everything but not sure how to do that cc @futurepaul

TonyGiorgio (approved):

PR #1208, opened 2024-06-11
fix sentry double init

futurepaul:


src/workers/walletWorker.ts
```diff
@@ -102,52 +102,63 @@ export async function setupMutinyWallet(
...
-                /^https:\/\/mutinywallet\.com/
-            ],
-
-            // Capture Replay for 10% of all sessions,
-            // plus 100% of sessions with an error
-            replaysSessionSampleRate: 0.1,
-            replaysOnErrorSampleRate: 1.0
-        });
+    try {
+        if (reportDiagnostics && sentryenv !== "") {
```
TonyGiorgio, line 106: Password is here too, do you not want to also do the same logic you do with the megastore?

TonyGiorgio (changes requested):

TonyGiorgio (approved):

PR #1231, opened 2024-06-24
Skip federation selector on startup

benthecarman:


TonyGiorgio (approved):

TonyGiorgio:
Should remove the helpful tip that shows "add federation" on the home page.

futurepaul (approved):

PR #1244, opened 2024-07-01
Display proper swap error

benthecarman:
Before we were showing this `swap_lightning.too_small` error, but this does not cover all the cases of errors we could get, instead just display the error we get from mutiny-wasm

### DIFF
--- a/src/routes/SwapLightning.tsx
+++ b/src/routes/SwapLightning.tsx
@@ -119,7 +119,7 @@ export function SwapLightning() {
         return feeSats() !== 0n;
     });
 
-    const [feeEstimateWarning, setFeeEstimateWarning] = createSignal<string>();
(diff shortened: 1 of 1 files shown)


src/routes/SwapLightning.tsx
```diff
@@ -146,7 +146,7 @@ export function SwapLightning() {
             setStage("preview");
         } catch (e) {
             console.error(e);
-            setFeeEstimateWarning(i18n.t("swap_lightning.too_small"));
+            setFeeEstimateWarning(eify(e));
```
futurepaul, line 149: can keep the existing functionality:

```suggestion
            const err = eify(e);
            if (err.message === "Satoshi amount is invalid") {
                setFeeEstimateWarning(
                    new Error(i18n.t("swap_lightning.too_small"))
                );
            } else {
                setFeeEstimateWarning(err);
            }
```

futurepaul (approved):

PR #1245, opened 2024-07-01
Change LSP url to olympus

benthecarman:


TonyGiorgio (commented):
Signet too?

TonyGiorgio:
The playwright tests are failing now, which seem to indicate some difference between the old LSP and the new one. That should probably be looked into before we switch over.

benthecarman:
> The playwright tests are failing now, which seem to indicate some difference between the old LSP and the new one. That should probably be looked into before we switch over.

was just things timing out, think I got it working now.

TonyGiorgio (approved):
Cool. Go ahead and change in cloudflare too, it'll go out on the next deploy.

benthecarman:
> Cool. Go ahead and change in cloudflare too, it'll go out on the next deploy.

done

PR #1253, opened 2024-07-03
Correctly display the currently in use LSP

benthecarman:
for https://github.com/MutinyWallet/mutiny-node/pull/1234


futurepaul (commented):
works for me, just need the new version of mutiny-node and we can get this in

without:

Screenshot 2024-07-03 at 1 01 43 PM

with:

Screenshot 2024-07-03 at 12 59 13 PM

futurepaul (approved):

PR #1263, opened 2024-07-12
Fedimint resync

benthecarman:
needs https://github.com/MutinyWallet/mutiny-node/pull/1239

to test you need to be in a federation and in safe mode, to enter safe mode to do so, go to `/?safe_mode=true`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "mutiny-wallet",
-    "version": "1.7.13",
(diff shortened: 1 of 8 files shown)


futurepaul (approved):
lgtm everything seems to work. added a button to get you into safe mode from the emergency kit.

only caveat is I've never actually made it through a whole rescan because it takes so long so idk what happens at the end but logic seems straightforward

PR #1266, opened 2024-07-16
fix: search template is missing translations references

martinbarilik:
Search template is missing i18n translations.


benthecarman (commented):
Can you put the English translations in the en.json file

martinbarilik:
It is my understanding that they are already present in the master, they just weren't used yet.

benthecarman (approved):

PR #1293, opened 2024-09-03
hide voltage lsp option

futurepaul:
this is a gross way to code this but felt the easiest to reason about so I went with it


TonyGiorgio (approved):
Works for me

